PR #17, opened 2016-03-24
added textarea and password with readmes and demos

rox163:
# fix


tests/dummy/app/pods/area/template.hbs
```diff
@@ -0,0 +1,137 @@
...
+        {{format-markdown "```handlebars
+{{frost-textarea
+  class='error'
+}}
+```"
+        }}
+      </div>
+    </div>
+    <div class="example">
+      <div class="title">Read-only</div>
```
laflower, line 98: I think we want to remove the readonly demo.

laflower:
textarea has a test in tests/unit that is not in this PR. Nothing big, it just verifies that the component has the frost-textarea className. Not sure if you ignored it on purpose or forgot.

sglanzer-fire:
@rox163 please work with @laflower to resolve any missed changes

PR #40, opened 2016-04-04
updated multi select checkbox size. fix margin issue for some demos.

rox163:
# fix

### DIFF
--- a/tests/dummy/app/pods/area/template.hbs
+++ b/tests/dummy/app/pods/area/template.hbs
@@ -1,4 +1,5 @@
 <div class="dummy-body textarea">
+  <div class="section-header">frost-textarea</div>
 
   <div class="section-title">Default</div>
   <hr>
--- a/tests/dummy/app/pods/area/template.hbs
+++ b/tests/dummy/app/pods/area/template.hbs
@@ -1,5 +1,4 @@
 <div class="dummy-body textarea">
-  <div class="section-header">frost-textarea</div>
 
   <div class="section-title">Default</div>
   <hr>
--- a/tests/dummy/app/pods/select/template.hbs
+++ b/tests/dummy/app/pods/select/template.hbs
@@ -1,5 +1,4 @@
 <div class="dummy-body select">
-  <div class="section-header">frost-select</div>
 
   <div class="section-title">Content</div>
   <hr>
--- a/tests/dummy/app/pods/select/template.hbs
+++ b/tests/dummy/app/pods/select/template.hbs
@@ -1,4 +1,5 @@
 <div class="dummy-body select">
+  <div class="section-header">frost-select</div>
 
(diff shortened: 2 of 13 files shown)


tests/dummy/app/styles/app.scss
```diff
@@ -135,7 +135,6 @@ hr {
       display: inline-block;
       font-size: $frost-font-s;
       margin-right: 20px;
-      margin-top: -5px;
```
vesper2000, line 138: All the demos look good without this?  I recall adding it for some reason.

vesper2000, line 138: Actually, that was there all along - I see where the problem was.  Looks good.

tests/dummy/app/pods/area/template.hbs
```diff
@@ -1,4 +1,5 @@
 <div class="dummy-body textarea">
+  <div class="section-header">frost-textarea</div>
```
sglanzer-fire: We don't want the section headers in the doc site, so unless there is code to remove these headers please exclude them

sglanzer-fire:
Please remove the headers prior to approval

tests/dummy/app/pods/area/template.hbs
rox163: oh dang. I saw it added to the frost-select demo page so I got excited.

vesper2000: The only reason select has the headers is that there are two sub components: frost-select and frost-multi-select.  I added it to distinguish the two.   I did not add update the other components for the same reason as Steve mentioned above - they are not necessary for the docs.  I am assuming we would keep the select headers.

tests/dummy/app/pods/select/template.hbs
```diff
@@ -1,5 +1,4 @@
 <div class="dummy-body select">
-  <div class="section-header">frost-select</div>
```
vesper2000, line 2: So this one needs to came back.

tests/dummy/app/pods/select/template.hbs
```diff
@@ -1,4 +1,5 @@
 <div class="dummy-body select">
+  <div class="section-header">frost-select</div>
```
vesper2000, line 2: Should be all good now.

tests/dummy/app/pods/select/template.hbs
rox163, line 2: Already done in the last commit

sglanzer-fire:
Approved - thankya

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/40/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #46, opened 2016-04-11
BPSO-16029 Remove explicit width from frost-text

ghost:
PR Review Task: https://agile-jira.ciena.com/browse/BPSO-16410
This is a #PATCH#

We want consumers to define widths of components.

### DIFF
--- a/addon/styles/_frost-text.scss
+++ b/addon/styles/_frost-text.scss
@@ -15,7 +15,6 @@ $padding: 0 19px 0 6px;
 .frost-text {
   display: flex;
   align-items: center;
-  width: 175px;
   &.error {
     > input {
       border: $error-border;


addon/styles/_frost-text.scss
```diff
@@ -15,7 +15,6 @@ $padding: 0 19px 0 6px;
 .frost-text {
   display: flex;
   align-items: center;
-  width: 175px;
```
job13er, line 18: @sglanzer I'm guessing we don't want to have explicit `width` either on components, right?

sglanzer-fire, line 18: It's debatable - in some cases you want a default, but it's really hard to pick a good default - 100%? 175px? 10em?  It really changes according to the consuming context.  In general I would expect text to primarily be used in bunsen forms, so bunsen should be responsible for the width.

So is there harm in providing a global default?  Probably not, as long as it doesn't restrict the override capabilities/make it hard to get what you want.

The consequence of this removal is going to end up with a default width (either px or 100% anyway), so it's probably better to keep it - thoughts?

job13er, line 18: I guess it doesn't prevent consuming context from overriding, but the same can be said for the `max-width` that was already removed, right? If you don't want to make people override that, why do you want to make people override this? Or is the argument that fewer people would want to override `width` than would want to override `max-width` or `min-width`?

job13er, line 18: Sorry for the late reply, I had to set up a rule to filter out my github emails, they were getting ridiculous. I need to figure out how to get just the ones I actually care about. Maybe a digest or something...

sglanzer-fire, line 18: Offline conversation:

@sglanzer To clarify my comment, if you remove the explicit `width`, the field will still ​_have_​ a width, it just won’t be explicit, so it’ll either go to browser default or it’ll go to 100% width.  Basically, I didn’t see any harm in making it our default

@job13er: ah, I see, so you're saying, chances are you'll need to override it regardless, so why not make it consistent if you don't

sglanzer-fire, line 18: Given the above notes, I'm going to close this PR

PR #48, opened 2016-04-13
Fix issue with checkbox not updating based on property

sandersky:
# PATCH

This makes `frost-checkbox` update when the `checkbox` property changes.

### DIFF
--- a/addon/components/frost-checkbox.js
+++ b/addon/components/frost-checkbox.js
@@ -7,10 +7,17 @@ export default Ember.Component.extend({
   classNames: ['frost-checkbox'],
   classNameBindings: ['size'],
 
-  didInitAttrs () {
-    const checked = this.attrs ? this.attrs.checked : false
-    this.set('isChecked', checked)
+  isChecked: Ember.computed('checked', function () {
+    const checked = this.get('checked')
 
+    if ([null, undefined, false].indexOf(checked) !== -1) {
+      return false
+    }
+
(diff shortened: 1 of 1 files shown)


addon/components/frost-checkbox.js
```diff
@@ -7,10 +7,17 @@ export default Ember.Component.extend({
   classNames: ['frost-checkbox'],
   classNameBindings: ['size'],
 
-  didInitAttrs () {
-    const checked = this.attrs ? this.attrs.checked : false
-    this.set('isChecked', checked)
+  isChecked: Ember.computed('checked', function () {
+    const checked = this.get('checked')
+
+    if ([null, undefined, false].indexOf(checked) !== -1) {
```
sglanzer-fire, line 13: I don't suppose there is a simpler version of a falsey check?

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/48/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-checkbox.js
sandersky, line 13: I suppose I could change this to:

``` javascript
@computed('checked')
isChecked (checked) {
  return [null, undefined, false].indexOf(checked) === -1
}
```

PR #49, opened 2016-04-13
adding ember-frost-scroll

vesper2000:
# patch

Still need to add README - will create an issue to track it.

### DIFF
--- a/addon/components/frost-scroll.js
+++ b/addon/components/frost-scroll.js
@@ -1,0 +1,25 @@
+import Ember from 'ember'
+import _ from 'lodash/lodash'
+
+const {Component, on, run} = Ember
+
+const scrollYEndContext = {
+  name: 'on-scroll-y-end'
+}
+const debouncePeriod = 150
+
+export default Component.extend({
+  classNames: ['frost-scroll'],
+
+  initializeScroll: on('didInsertElement', function () {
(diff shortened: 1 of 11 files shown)


sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/49/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
This should go in as a #feature# / #minor# - nevermind, we're still in pre-1.0 - carry on ;)

sandersky:
@sglanzer This PR has me wondering if `ember-frost-core` were already on `1.0` would this need to be a major bump? I ask because the consumer must have `perfect-scrollbar` in their `bower.json` file and if they have `"ember-frost-core": "^1.0.0"` in their `package.json` this PR would have caused their build to break (given this were a minor).

sglanzer-fire:
@sandersky That's a legit point and I think it cycles back to our concerns over dependency upgrades - long story short, I believe you're right - this would technically be a breaking change.  We could potentially avoid that by using the npm publish of perfect-scrollbar https://www.npmjs.com/package/perfect-scrollbar and using browserify, which is probably what we should do to avoid an additional bower dependency

PR #52, opened 2016-04-13
fix height setup

rox163:
# fix

to avoid overriding any liquid containers that are first children of `<body>`
'setupHeight' is a custom class that can be applied by consumers to their app containers in case they have an obscure setup that doesn't fall in this rule.

### DIFF
--- a/addon/styles/_flex.scss
+++ b/addon/styles/_flex.scss
@@ -1,6 +1,6 @@
 html,
 body,
-body > .ember-view {
+body > .ember-view:not(.liquid-target-container), .setupHeight.ember-view {
   height: 100%; // 100% height is used for vertical flexbox
 }
 
@@ -9,3 +9,4 @@ body > .ember-view {
     flex: (#{$x});
   }
 }
+


sglanzer-fire:
I wouldn't mind a quick chat on this, I don't really get what's going on with setupHeight
1.  If we can provide a custom class on the app container, why not just use that to set the 100% height
2.  I don't get what that class is supposed to provide to non-standard consumers since it applies the same rule

sglanzer-fire:
Or are you saying that the Ember container element may be something embedded in a wider app that could be targeted for those consumers?

sglanzer-fire:
After poking at this for a little, you could also go this route

```
import Ember from 'ember';

export default Ember.Route.extend({
  addRootClass: Ember.on('init', function() {
    Ember.run.schedule("afterRender", this, function() {
      Ember.$(`:first .ember-view`).addClass('test')
    });
  })
});
```

Which gives us a handle on the root Ember element and lets us use a non parent/child selector

sglanzer-fire:
I think this is a better approach, but let me know what you think

rox163:
I dont think Ember.$(`:first .ember-view`).addClass('test') would work because the first ember-view is the liquid-tether. 
Maybe my css is a bit wrong? I was trying to provide the same style as a separate class that consumers could use in case the `not(:liquid-target-container)` rule didnt apply to their situation [some other view was the first one and not the app] and they needed the `height:100%` style applied to their app container. So that class would be something we document that they could use. Or just ask them to apply the height style themselves.

sglanzer-fire:
Your CSS is fine - my only concern is that the negation can be brittle - e.g. if we import a different addon or they change the target class then we're broken again.  If we can manage to set up a positive class that targets the top ember view container, then we're safe.

I was trying to mess around with the Ember container and views to find the right element to add the class to so that we could take that approach.

I'll add an issue for that approach and we'll merge this

sglanzer-fire:
Raised https://github.com/ciena-frost/ember-frost-core/issues/61

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/52/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #56, opened 2016-04-15
Textarea blur

TheAxeMan301:
# PATCH

Previously an onFocus attribute to a frost-text would overwrite its internal handler that selects the text. Most other widgets did not support onFocus. Fix that and add it to the demo. Can't figure out how to test it though.
Also set tabIndex=-1 so that tabbing skips the textarea clear button and the password reveal button.
@sglanzer @sandersky

### DIFF
--- a/addon/components/frost-button.js
+++ b/addon/components/frost-button.js
@@ -93,6 +93,7 @@ export default Ember.Component.extend({
     'autofocus',
     'disabled',
     'type',
+    'tabIndex',
     'title'
   ],
 
@@ -221,6 +222,13 @@ export default Ember.Component.extend({
 
     if (!this.get('disabled') && _.isFunction(this.attrs['onClick'])) {
       this.attrs['onClick'](this.get('id'))
+    }
+  }),
+
+  _onFocus: Ember.on('focusIn', function (e) {
+    // If an onFocus handler is defined, call it
+    if (this.attrs.onFocus) {
+      this.attrs.onFocus()
     }
   })
 })
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 
 # misc
 /.sass-cache
+/.idea
 /connect.lock
 /coverage/*
 /libpeerconnection.log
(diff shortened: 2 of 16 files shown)


sglanzer-fire:
Thanks for this @TheAxeMan301 - I'm happy to merge this and raise an issue to address the testing

sglanzer-fire:
Added #57 to track the missing tests

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/56/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-button.js
```diff
@@ -222,5 +223,12 @@ export default Ember.Component.extend({
     if (!this.get('disabled') && _.isFunction(this.attrs['onClick'])) {
       this.attrs['onClick'](this.get('id'))
     }
+  }),
+
+  _onFocus: Ember.on('focusIn', function (e) {
+    // If an onFocus handler is defined, call it
+    if (this.attrs.onFocus) {
+      this.attrs.onFocus()
+    }
```
sandersky, line 232: I find defining arbitrary keys like this odd when they are not used and personally think it would be better to define this in the `init` method. Also it would probably make sense to pass the argument to the consmuer `onFocus` property as well.

``` javascript
init () {
  this._super(...arguments)

  Ember.on('focusIn', (e) => {
    const onFocus = this.get('onFocus')

    if (onFocus) {
      onFocus(e)
    }
  })
}
```

Also by putting it in the `init` method you have better encapsulation as the consumer could still technically pass in a property `_onFocus` that overrides the one you've defined.

sandersky:
Overall nice work!

addon/components/frost-button.js
sglanzer-fire, line 232: @sandersky Good points - I was also wondering if we could alias in on top of the Ember events - e.g. map `onFocusIn` (our event format) to `focus-in` (Ember's event format) - might be something we could mess with globally via a Component re-open so that we could remap all of the events in one shot

PR #65, opened 2016-04-20
adding new button design

vesper2000:
# patch

### DIFF
--- a/addon/styles/_frost-button.scss
+++ b/addon/styles/_frost-button.scss
@@ -27,14 +27,6 @@ $frost-color-button-info-bar-active:  rgba($frost-color-blue-1, 0.1);
 $frost-color-button-in-line-darken-1: darken($frost-color-blue-1, 8);
 $frost-color-button-in-line-darken-2: darken($frost-color-blue-1, 15);
 
-@mixin button-color ($color) {
-  color: $color;
-  svg * {
-    stroke: $color;
-    fill: $color;
-  }
-}
-
 .frost-button {
   padding: 0;
   border: none;
@@ -67,7 +59,7 @@ $frost-color-button-in-line-darken-2: darken($frost-color-blue-1, 15);
   // priority = 'primary'
   &.primary {
     background: $frost-color-button-primary-bg;
-    @include button-color($frost-color-button-primary);
+    color: $frost-color-button-primary;
 
     &:hover:not(.disabled) {
       background: $frost-color-button-primary-hover-bg;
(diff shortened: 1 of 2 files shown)


job13er:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/65/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/styles/_frost-button.scss
```diff
@@ -305,6 +297,54 @@ $frost-color-button-in-line-darken-2: darken($frost-color-blue-1, 15);
       box-shadow: none;
     }
   }
+  
+  // design = 'app-bar'
+  &.app-bar {
```
sglanzer-fire, line 302: Is this unique in any way other than the background color from a standard icon + text?  If not, then I'd prefer that this be included from the context where the design will be used - e.g. Design is for global common layouts and the one-offs are included from the unique context

job13er, line 302: Sorry, I'm not sure I follow what that means.

> I'd prefer that this be included from the context where the design will be used

Do you mean the app-bar itself should have these styles? 

I thought we had this argument in slack and you were pitching for having all this in the button itself...

> e.g. Design is for global common layouts and the one-offs are included from the unique context

I thought one-offs were exactly what `design` was for. It appears I may be out-of-the-loop here. Sorry, I guess I shouldn't have approved this. I'll refrain next time.

vesper2000, line 302: They have their own look and there are no effects for this kinds of button.
I thought it fit well for the design type as it did not have any other required parameters.

sglanzer-fire, line 302: Ok cool, I forgot that the intention of "design" was to avoid this type of debate over placement and make refactoring common parts easier - thanks for the reminder, carry on and ignore me

vesper2000, line 302: Just as a side note, the app bar is nice and clean now, each item is defined as follows:

```
 {{frost-button
      class='navigation'
      design='app-bar'
      icon='icons/app/nav-network'
      text=(t 'application-bar.network')
      onClick=(action 'openNetworkNavigation')
    }}
```

And it used to be like this:

```
   {{#frost-button class='navigation' on-click=(action 'openNetworkNavigation')}}
        <div class='icon-text'>
          {{frost-svg path='app/nav-network'}}
            <div class='text'>{{t 'application-bar.network'}}</div>
        </div>
    {{/frost-button}}
```

PR #85, opened 2016-04-29
Fix issue with selectedValue not changing but data changing

sandersky:
# PATCH

If the `selectedValue` property is set but `data` is empty, then `data` becomes set, the select doesn't update to actually select the item with the `selectedValue`.


job13er:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/85/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #108, opened 2016-06-06
Fixing the icon pack deprecation notice

sglanzer-fire:
# fix# turns out the deprecation is shown if the value is falsey, so the test had to be reversed.  The deprecation should now read as - "we're good if the path does **not** include `/`, otherwise display the notice"

Also adding default fills to some of the frost icons to make the coloring flexible (and demonstrate how to set the fill color)
# CHANGELOG

Fixed the deprecation notice test for nested icon paths
Added CSS based default fills to some icons from the frost icon pack
Fixed broken checkbox tests


addon/components/frost-icon.js
```diff
@@ -17,7 +17,7 @@ export default Component.extend({
   didReceiveAttrs ({newAttrs}) {
     deprecate(
       'nested icon paths have been deprecated in favor of flat icon packs',
-      _.includes(newAttrs.icon, '/'),
+      !_.includes(newAttrs.icon.value, '/'),
```
sandersky:
``` js
!_.includes(_.get(newAttrs, 'icon.value'), '/'),
```

sglanzer-fire: Thanks, null guards are 👍

sandersky:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/108/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #110, opened 2016-06-06
Rollback to 49f7b92ba2ba703378914c6ef4a7f416a6d58a24 for FrostCheckboxComponent

srowhani:
# fix


sglanzer-fire:
We managed to fix the tests (selector issue), so we don't need the rollback

PR #117, opened 2016-06-15
Start cleaning up code

sandersky:
# PATCH
# CHANGELOG
- **Added** `unit: true` to all unit tests to explicitly mark them as unit tests.
- **Moved** templates to `app/` directory so consumers can override them if they wish.
- Started using destructuring for Ember properties such as `const {Component} = Ember` instead of using `Ember.Component`.


sglanzer-fire:
Approved - thanks for the cleanup, you rule :)

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/117/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #118, opened 2016-06-15
More cleanup

sandersky:
# PATCH
# CHANGELOG
- **Added** section comments in components for _dependencies_, _properties_, _computed properties_, _functions_, _events_, and _actions_.
- **Fixed** bug where disabled text input with text could be cleared as clear icon was present when it shouldn't be.
- **Updated** components to use `ember-computed-decorators`.
- **Updated** `frost-icon` and `frost-link` to use `ember-prop-types`.


addon/components/frost-icon.js
```diff
@@ -1,22 +1,54 @@
-import Ember from 'ember'
-import { PropTypes } from 'ember-prop-types'
 import _ from 'lodash'
+import Ember from 'ember'
+const {Component, deprecate} = Ember
+import computed, {readOnly} from 'ember-computed-decorators'
+import PropTypeMixin, {PropTypes} from 'ember-prop-types'
+
+export default Component.extend(PropTypeMixin, {
```
sglanzer-fire, line 7: Are we _positive_ the mixin is necessary?  I haven't seen any adverse affects from just relying on the initializers

tests/dummy/app/pods/icons/controller.js
```diff
@@ -1,6 +1,6 @@
+import _ from 'lodash'
```
sglanzer-fire, line 1: I don't believe we've recorded the preferred ordering of imports (I had been ordering alphabetically on the package being imported from) - probably worth stating in the style guidelines.

Is this something we could eslint --fix?

sglanzer-fire:
I'm a fan of section comments, but empty sections always feel unnecessary to me.  I suppose they aren't if you add new code and it's something I can live with, but I would be happier if we stripped out empty sections.

Alternatively, I've always wanted to see if this sort of thing could have better treatment in an editor, might be worth looking into for Atom and VSC

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/118/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-icon.js
sandersky, line 7: I believe I had run into issues with `getDefaultProps()` not working properly without the mixin. Maybe it was unrelated but I'm not sure.

tests/dummy/app/pods/icons/controller.js
sandersky, line 1: I'd love an eslint rule to address this.

PR #122, opened 2016-06-15
Fix frost-select not clearing

sandersky:
# PATCH
# CHANGELOG
- **Fixed** issue with `frost-select` and consumer clearing `selectedValue` but DOM not clearing input.

### DIFF
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -266,16 +266,21 @@ export default Component.extend({
   },
 
   /* Ember.Component method */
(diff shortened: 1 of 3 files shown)


sglanzer-fire:
Approved - I'm looking into the DOM cleanup issues with remodal to see if we can address that as well

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/122/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #129, opened 2016-06-22
add protection icons

ghost:
# MINOR
# CHANGELOG
- added shield-full, shield-half icons
  shield-full
  shield-half


sglanzer-fire:
I think these belong within a Ciena internal icon-pack.  We either need to create this project (feel free to take that initiative) or add them to the applicable app for now.

rox163:
FYI
image

ghost:
i_protection_full

sglanzer-fire:
Thanks for the images, I suppose we could include them in `core` if they were named generically (shield, half-shield?)

I would still recommend moving them into an internal icon pack, or barring that an app.

ghost:
No problem.  I will let @sandersky respond as to where to ultimately put these, but for now I have put them in the app that needs them.  thx.

sandersky:
If these remain in the frost icon pack then I agree with @sglanzer as to make the names more generic. Shield icons are found in various free icon packs so I see no harm in including them here to provide more options to consumers of the frost icon pack.

sglanzer-fire:
I'm good with a rename

ghost:
It turns out there already is an icon named  `shield.svg`  which is similar to (but not identical to) my `protection-partial`
shield
So what would you like my icons to be named?

sglanzer-fire:
Let's go with shield-full.svg and shield-half.svg and we can see about removal of the old shield.svg down the line

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/129/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #131, opened 2016-06-24
Add key up/down events for frost-text

job13er:
This #feature# adds properties to `frost-text` to allow consumers to listen in on the
`keyUp` and `keyDown` events on the `input` element within `frost-text`

I also moved the `.md` files for component documentation under a `docs/` directory to clean up the top-level of the repo a little.
# CHANGELOG
- **Added** `onKeyUp` property on `frost-text` that receives `keyUp` events from the inner `input`
- **Added** `onKeyDown` property on `frost-text` that receives `keyDown` events from the inner `input`


sglanzer-fire:
I'm working on a PR to expose events with Frost naming in a more programatic style.  I don't think it will conflict, but I need to put a little bit more work into it before I'm sure.

I agree with you that the onInput event should be pure (event.target.id, event.target.value) - I suppose we could transition to this at least in part by putting `target` as a property on the current object and warning that we'll switch entirely to the `event` object in the future - thoughts?

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/131/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

job13er:
> I'm working on a PR to expose events with Frost naming in a more programatic style. I don't think it will conflict, but I need to put a little bit more work into it before I'm sure.

👍 The thought of manually having to add all of them as we want to use them was a bit horrifying 😱 

> I agree with you that the onInput event should be pure (event.target.id, event.target.value) - I suppose we could transition to this at least in part by putting target as a property on the current object and warning that we'll switch entirely to the event object in the future - thoughts?

I like that approach. We could even provide the whole `event` in the current object, so it becomes

``` js
{
  event: {},
  id: 'foo',
  value: 'bar'
}
```

Then warn that in the next major version, the object will be replaced with just the `event` object itself. 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/131/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
Great - I'll go that route with the next PR

PR #132, opened 2016-06-30
Inline svg config option

EWhite613:
# minor

### DIFF
--- a/addon/instance-initializers/svg-use-polyfill.js
+++ b/addon/instance-initializers/svg-use-polyfill.js
@@ -1,7 +1,8 @@
 import svg4everybody from 'npm:svg4everybody'
-import config from 'ember-get-config'
+import Ember from 'ember'
+const {ENV} = Ember
 export function initialize () {
-  if (config.iconPacks && config.iconPacks.inline) {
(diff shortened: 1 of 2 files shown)


addon/instance-initializers/svg-use-polyfill.js
```diff
@@ -1,7 +1,15 @@
 import svg4everybody from 'npm:svg4everybody'
-
+import Ember from 'ember'
+const {ENV} = Ember
 export function initialize () {
-  svg4everybody()
+  if (ENV.iconPacks && ENV.iconPacks.inline) {
```
sandersky, line 5: Would be a good place to use lodash:

``` js
if (_.get(ENV, 'iconPacks.inline')) {
  …
```

sglanzer-fire:
Approved - move to a minor release

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/132/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
I'd like to see a comment in the README as well about this option and the circumstances where it could be useful.

PR #135, opened 2016-06-30
Fixing select to avoid double firing

sglanzer-fire:
# fix

The select by value is only applicable when receiving attribute changes from an external source and was firing the onChange twice on user input
# CHANGELOG
- Changed selection by value to not fire the onChange event


sandersky:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/135/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #141, opened 2016-07-11
Created blueprint for generic frost-component

srowhani:
# patch


srowhani:
File generated from `ember g frost-component test-component`

``` js
import Ember from 'ember'
import FrostEvents from 'ember-frost-core/mixins/frost-events'
import {
  PropTypes
} from 'ember-prop-types'

const {
  Component
} = Ember

import layout from '../templates/components/test-component';

export default Component.extend(FrostEvents, {
    // == Properties =============================================================
    excludeEvents: [],
    classNames: [
        'test-component'
    ],
    layout,
    propTypes: {},
    // == Functions =============================================================
    getDefaultProps () {
        return {}
    }
})
```

sandersky:
Can we potentially override the default blueprint for components instead? I'd much prefer to do `ember g component test-component` to `ember g frost-component test-component`? This would also reduce the Ciena/Frost specific knowledge required by developers.

srowhani:
It can be done. @sglanzer thoughts?

sglanzer-fire:
Yeah, I'm cool with that - based on the discussions yesterday I think it also makes sense to make the blueprints a separate project (this would also allow us to install that project globally to allow repo inits)

sglanzer-fire:
https://github.com/ciena-frost/ember-frost-blueprints

PR #144, opened 2016-07-19
Cleanup tests and fix bug in frost-select

job13er:
This is a #fix# 

There was an `.eslintrc` in `tests/` which was defining some globals
which really shouldn't have been globals, so that was removed and the
appropriate `import` statements were added to affected tests.

There was also a bug stopping `frost-select` from allowing users to
type queries if the `selectedValue` was passed in but null
(which happens to be how `ember-frost-bunsen` uses it).

`eslint` was also updated to `3.x`
# CHANGELOG
- **Fixes** a bug in `frost-select` where the options popover was closed whenever someone typed into the input (when `selectedValue` is passed in as props).
- **Updated** `eslint` within the project to `3.x`
- **Added** some linting rules for `mocha` (via an update to `eslint-config-frost-standard`)
- **Updated** `tests/` directory to no longer specify an `.eslintrc` with `globals` that shouldn't actually be globals.


addon/components/frost-select.js
```diff
@@ -270,15 +270,14 @@ export default Component.extend({
...
-
     const dataChanged = isAttrDifferent(newAttrs, oldAttrs, 'data')
     const selectedChanged = isAttrDifferent(newAttrs, oldAttrs, 'selected')
     const selectedValueChanged = isAttrDifferent(newAttrs, oldAttrs, 'selectedValue')
 
     // If frost-select instance is being reused by consumer but context is cleared make
     // make sure to actually clear input (noticed when used in conjunction with dialog
     // compoonents that don't destroy DOM when closed and re-opened)
-    if ('selectedValue' in newAttrs && newAttrs.selectedValue.value === undefined) {
+    if (selectedValueChanged && ('selectedValue' in newAttrs) && (newAttrs.selectedValue.value === undefined)) {
```
job13er, line 280: Here's the actual bugfix (may be hard to find with all the linting updates)

tests/dummy/app/pods/select/template.hbs
```diff
@@ -172,6 +172,7 @@
       <div class="demo">
         {{frost-select
           data=data
+          selectedValue=externalSelectedValue
```
job13er, line 175: This was added just so that the demo exercises the same code path that bunsen will.

tests/helpers/module-for-acceptance.js
```diff
@@ -1,23 +0,0 @@
-import { module } from 'qunit'
```
job13er, line 1: This was hanging around from when the addon was scaffolded, I think.

tests/integration/components/frost-select-test.js
```diff
@@ -92,78 +93,67 @@ describeComponent(
     })
 
     it('lists all passed data records', function () {
-      run.later(() => {
-        expect(this.$('.frost-select li').length).to.eql(props.data.length)
-      })
+      expect(this.$('.frost-select li').length).to.eql(props.data.length)
```
job13er, line 96: This file has a lot of these, basically I just swapped code that looked like this:

``` js
// do something
run.later(() => {
  // check something
})
```

To look like this: 

``` js
run(() => {
  // do something
})
// check something
```

Making sure all the `// do something` happen in a run-loop instead of waiting till the next run loop to `// check something`

tests/integration/components/frost-select-test.js
```diff
@@ -182,173 +172,147 @@ describeComponent(
...
-            value: 'Tony Starks',
-            label: 'Ghostface'
-          }
-        ],
-        greeting: 'Hola'
-      }
-      this.setProperties(props)
-      this.render(selectedValueTestTemplate)
-    })
+    describe('when passing in selected value', function () {
```
job13er, line 278: Oh yeah, there were two `describeComponent` calls in the one module, which seemed off to me.  Since they were for the same component anyway, I just made the latter into a `describe()` within the first `describeComponent`

tests/unit/components/frost-select-test.js
```diff
@@ -98,10 +99,39 @@ describeComponent(
           })
         })
 
-        it('sets selected to expected', function () {
+        it('should call selectOptionByValue()', function () {
           expect(component.selectOptionByValue.lastCall.args).to.eql(['foo'])
         })
       })
+
+      describe('when no selected value', function () {
```
job13er, line 107: Here's the actual added test for the bugfix (fails w/o the fix)

index.js
```diff
@@ -91,13 +92,15 @@ module.exports = {
     // No icon pack path defined and an app = legacy case where the app icons must be merged into the 'frost' icon pack
     const isLegacy = !_.has(this, 'app.options.iconPackOptions.path') && !isAddon
 
-    if (isLegacy && fs.existsSync(path.join(this.project.root, 'public/svgs'))) {
-      iconNames['frost'] = _.concat(iconNames['frost'], this.flattenIcons([], '', path.join(this.project.root, 'public/svgs')))
+    const svgPath = path.join(this.project.root, 'publich/svgs')
```
sglanzer-fire: publich

sglanzer-fire:
Didn't realize testem could be done as a module, very nice.  Also going to steal the eslint test globals for the new scaffolding.

Do we have a convention on when to line wrap destructured imports and array properties?  I've been line breaking on all entries.

sglanzer-fire:
Once publich is fixed - approved

job13er:
> Do we have a convention on when to line wrap destructured imports and array properties? I've been line breaking on all entries.

Not that I know of, but we probably should. I generally don't line-break them until they become too long for a single line, then I make it one per line. Not sure if that's the way to go, but it's what I've been doing.

index.js
job13er: The less commonly acknowledged spelling of `public` it's a silent `h` at the end ;)

job13er:
> Didn't realize testem could be done as a module

I didn't either until @sandersky pointed out that new ember generated projects use the `.js`

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/144/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #147, opened 2016-07-20
Updated behavior/events for text/password

sglanzer-fire:
# minor

I need to move on to a few other high priority items, so as long as it's acceptable to everyone I'll raise issues to add tests and add propTypes for text/password
# CHANGELOG
## Text/Password native event support
- Mapped all available DOM events to Frost events (e.g. focusIn -> onFocusIn) see the demo for a full list
- Event hooks respond with native DOM events instead of {id, value} object, this is still supported as a legacy usage, but please switch (no way to detect when the legacy attributes are used unfortunately)
- Deprecated `onFocus` (replaced by `onFocusIn`)
- Deprecated `onBlur` (replaced by `onFocusOut`)
- Removed the `excludedEvents` property from the FrostEvents mixin, this is covered automatically when local events are present
## Text/Password visual and behavioral updates
- Added a new color (grey-7) to the palette for input borders
- Updated the visuals and animation for hover, focus, clear and show/hide
## Text/Password demo updates
- Demos demonstrate legacy, native, text support and deprecated events
- Updated the text/password demos to use snippets
## New keycodes util
- Added a keycodes enum for use in matching keycodes from keyboard events
## Code cleanup
- Integrated ember-concurrency for async behavior in text/password (clear text, show/hide password)
- Text/password actions are fully run-loop compliant and clean up automatically on component destruction
## Future
- Moving towards dropping ember-one-way-controls from the dependency list (requires Ember 2.3.1+)


addon/components/frost-password.js
```diff
@@ -1,103 +1,82 @@
...
+  },
 
-  // ==========================================================================
-  // Events
-  // ==========================================================================
+  // == Events ================================================================
 
-  focusOut: Ember.on('focusOut', function () {
+  _focusOut: on('focusOut', function (event) {
+    // this._super(...arguments)
```
sandersky: Remove commented line.

addon/components/frost-password.js
```diff
@@ -1,103 +1,82 @@
...
-  }),
-
-  keyPressed: Ember.on('keyPress', function (e) {
-    var s = String.fromCharCode(e.which || e.keyCode)   // IE support
-    if ((s.toUpperCase() === s && s.toLowerCase() !== s && !e.shiftKey) ||
-      (s.toUpperCase() !== s && s.toLowerCase() === s && e.shiftKey)) { // caps is on
+  _keyPress: Ember.on('keyPress', function (event) {
+    const c = String.fromCharCode(event.which || event.keyCode)
+    if ((c.toUpperCase() === c && c.toLowerCase() !== c && !event.shiftKey) ||
+      (c.toUpperCase() !== c && c.toLowerCase() === c && event.shiftKey)) {
```
sandersky: Can we refactor this to something like below to make it more clear what you are doing:

``` js
  _keyPress: Ember.on('keyPress', function (event) {
    const c = String.fromCharCode(event.which || event.keyCode)
    const userInputIsUppercase = (c.toUpperCase() === c && c.toLowerCase() !== c)
    const userInputIsLowercase = (c.toUpperCase() !== c && c.toLowerCase() === c)
    const userIsHoldingShiftKey = event.shiftKey

    // If the user is not holding down the shift key and we get an uppercase character then caps lock is enabled.
    // Caps lock is also enabled if the user is holding down the shift key and we get a lowercase character.
    if (
      (userInputIsUppercase && !userIsHoldingShiftKey) ||
      (userInputIsLowercase && userIsHoldingShiftKey)
    ) {
      this.set('isCapsOn', true)
    }
  }),
```

addon/components/frost-password.js
```diff
@@ -1,103 +1,82 @@
...
+  _focusOut: on('focusOut', function (event) {
+    // this._super(...arguments)
     this.set('isCapsOn', false)
   }),
 
-  keyDown: Ember.on('keyDown', function (e) {
+  _keyDown: Ember.on('keyDown', function (e) {
     var s = e || window.e
-    if (this.get('isCapsOn') === false && (s.which === 20 || s.keyCode === 20)) {
+    if (s.which === 20 || s.keyCode === 20) {
```
sandersky: Rather than using `20` throughout this file can you define a constant at the top of the file and reference that.

``` js
const CAPS_LOCK_KEY = 20

…

    if (s.which === CAPS_LOCK_KEY || s.keyCode === CAPS_LOCK_KEY) {
…
```

addon/components/frost-password.js
```diff
@@ -1,103 +1,82 @@
...
     toggleReveal () {
-      this.toggleProperty('revealed')
-      this.set('type', this.get('revealed') ? 'text' : 'password')
-      this.set('revealIcon', this.get('revealed') ? 'hide' : 'show')
+      this.toggleProperty('isRevealed')
+      const $input = this.$('input').get(0)
+      $input.focus()
+      // Move the cursor to the end of the input
+      run.next(this, function () {
+        $input.selectionStart = $input.selectionEnd = 10000
```
sandersky: Since `$input` is referenced in a `run.next()` call I think it is possible for the DOM to already be destroyed and thus we probably need a sanity check here.

``` js
run.next(this, function () {
  if (document.contains($input[0])) {
    $input.selectionStart = $input.selectionEnd = 10000
  }
})
```

addon/components/frost-text.js
```diff
@@ -1,51 +1,105 @@
...
+      this.setProperties({
+        isClearEnabled: true,
+        isClearVisible: true
       })
-    })
-  },
+    } else {
+      this.set('isClearVisible', false)
+      yield timeout(200)
+      this.set('isClearEnabled', false)
```
sandersky: What is going on here exactly? Are we hiding the clear button then .2s later we are disabling it?

addon/components/frost-text.js
```diff
@@ -1,51 +1,105 @@
...
+  // Setting 'keyUp' directly on the {{input}} helper overwrites
+  // Ember's TextSupport keyUp property, which means that other
+  // TextSupport events (i.e. 'enter' and 'escape') don't fire.
+  // To avoid this, we use the TextSupport 'key-up' event and
+  // proxy the event to the keyUp handler.
+  actions: {
+    clear () {
+      this.$('input').focus()
+      this.$('input').val('')
+      this.$('input').trigger('input')
```
sandersky: It seems inefficient to lookup the same element 3 times in a row.

``` js
this.$('input')
  .focus()
  .val('')
  .trigger('input')
```

addon/templates/components/frost-password.hbs
```diff
@@ -1,28 +1,62 @@
...
+  onMouseMove=_eventProxy.mouseMove
+  onMouseUp=_eventProxy.mouseUp
+  onSubmit=_eventProxy.submit
+  onTouchCancel=_eventProxy.touchCancel
+  onTouchEnd=_eventProxy.touchEnd
+  onTouchMove=_eventProxy.touchMove
+  onTouchStart=_eventProxy.touchStart
+
+  onEnter=_eventProxy.enter
+  onEscape=_eventProxy.escape-press
```
sandersky, line 45: I think you covered them all! 😝

addon/templates/components/frost-text.hbs
```diff
@@ -1,14 +1,159 @@
...
+  touchStart=_eventProxy.touchStart
+
+  escape-press=_eventProxy.escape-press
+  insert-newline=_eventProxy.enter
+}}
+{{frost-text-clear
+  onClick=(action 'clear')
 }}
+
+{{!--
```
sandersky: What's with the commented out Javascript in this HTMLBars file?

addon/components/frost-password.js
sandersky: Also would it make sense to add `&& !this.get('isCapsOn')` to the `if` check so we don't set the property when it didn't actually change?

addon/components/frost-password.js
sglanzer-fire: Yeah, I wanted to handle this via ember-concurrency, but for some reason the property wouldn't take effect - I'll add the guards

addon/components/frost-text.js
sglanzer-fire: I'll add comments - focus events are a pain.  Basically we get the focusOut on the input prior to the click on the clear, which gets processed and hides the clear, so it's not available to click on.  This prolongs the clickability of the clear until it's fully transparent.

addon/templates/components/frost-password.hbs
sglanzer-fire, line 45: Any comments on the use of `eventVersion` for backwards compatibility?  I still need to add a deprecation notice, but that was the plan until 1.0

addon/templates/components/frost-text.hbs
sglanzer-fire: Whoop, copy/paste miss

package.json
```diff
@@ -112,6 +112,7 @@
     "ember-cli-uglify": "^1.2.0",
     "ember-code-snippet": "1.3.0",
     "ember-computed-decorators": ">=0.2.2 <2.0.0",
+    "ember-concurrency": "0.7.8",
```
sglanzer: I need to add this to the blueprint

tests/dummy/app/pods/field/template.hbs
```diff
@@ -12,7 +12,7 @@
     <div class="section">
       <div class="example">
         <div class="demo">
-          {{frost-text}}
+          {{frost-text value=model.default.value onInput=(action (mut model.default.value))}}
```
sglanzer-fire: Need to move these to use snippets as well

addon/templates/components/frost-password.hbs
sandersky, line 45: I think that works, I'd be tempted to more more rebellious and not make it backwards compatible but I guess that'd make me an 🐴 (closest emoji to a donkey I can find).

addon/components/frost-password.js
sglanzer-fire: Doesn't hurt (in case someone is using observers)

addon/templates/components/frost-password.hbs
sglanzer-fire, line 45: Added `id` and `value` to the root of the `event` object instead

sandersky:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/147/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #153, opened 2016-07-27
Adding ember-hook to core components

rox163:
# feature
## CHANGELOG
- **Added** Consumers can now use 'hook' attr of core components to enable easier testing with addons like ember-cli-page-object

### DIFF
--- a/addon/templates/components/frost-password.hbs
+++ b/addon/templates/components/frost-password.hbs
@@ -3,6 +3,7 @@
   class='frost-password-input'
   disabled=disabled
   form=form
+  hook=(concat hook '-text')
   maxlength=maxlength
   placeholder=placeholder
   readonly=readonly
@@ -45,7 +46,7 @@
   onEscape=_eventProxy.escape-press
 }}
 {{#if revealable}}
-  <div class='frost-password-reveal' onclick={{action (action 'toggleReveal')}}>
+  <div class='frost-password-reveal' data-test={{hook (concat hook '-reveal')}} onclick={{action (action 'toggleReveal')}}>
     {{revealMessage}}
   </div>
 {{/if}}
--- a/addon/templates/components/frost-password.hbs
+++ b/addon/templates/components/frost-password.hbs
@@ -3,7 +3,8 @@
   class='frost-password-input'
   disabled=disabled
   form=form
-  hook=(concat hook '-text')
+  hook=hook
+  isHookEmbedded=true
   maxlength=maxlength
   placeholder=placeholder
   readonly=readonly
--- a/addon/templates/components/frost-textarea.hbs
+++ b/addon/templates/components/frost-textarea.hbs
@@ -4,6 +4,7 @@
   disabled=disabled
   focus-out=(action 'onBlur')
   form=form
+  hook=(concat hook '-input')
   value=value
   placeholder=placeholder
   readonly=readonly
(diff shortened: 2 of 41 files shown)


juwara0:
👍 

We should also add a note to the README stating that `ember-hook` is now available on these components and list it as part of the API.

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/153/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

juwara0:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/153/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
I'd like to add a representative test for each hook to demonstrate that the targets are correct.  Some of our components are essentially shims around inputs that include additional targets, so it would be cool to see hooks on each of the sub-elements.  E.g. if a `frost-password` had hook='my-password' then it would be nice to see `my-password-input` hook, `my-password-reveal` hook and `my-password-clear`

I don't know if you want to address that in this PR, but it's something I definitely want

rox163:
Definitely agree with the test, otherwise just adding a hook attr doesnt show much. Maybe I can add one or two as examples for now.
Your password example though made me realize I missed some internal hooks. Right now its just on the outer component for something like that. Need to look at adding hooks on the reveal button and clear button for example.

rox163:
added a few tests in the latest commit as examples. Continuing to add tests to the other components

tests/integration/components/frost-button-test-.js
```diff
@@ -2,6 +2,7 @@ import {expect} from 'chai'
 import {describeComponent, it} from 'ember-mocha'
 import hbs from 'htmlbars-inline-precompile'
 import {beforeEach} from 'mocha'
+import {$hook, initialize} from 'ember-hook'
 
 describeComponent(
   'frost-button',
```
rox163, line 8: for some reason frost button integration tests are not running. all other component tests are running. Thats why this one is sparse for now.

EWhite613:
@rox163 looks like the button integration file ends with `-test-` rather than `-test`. Which would explain why it didn't run

rox163:
AH good catch!

rox163:
fixed broken tests and added hooks to internal components of frost-password

juwara0:
This looks good to me.

addon/templates/components/frost-password.hbs
```diff
@@ -3,6 +3,7 @@
   class='frost-password-input'
   disabled=disabled
   form=form
+  hook=(concat hook '-text')
```
sglanzer-fire: Can you make this `-input` instead?

sglanzer-fire: Actually, is that even necessary or do you just need to pass `hook=hook` and let `frost-text` append the `-input`?

rox163: Frost-text adds ‘-input’ though to its input field. So internally in password you can get to it with <hook>-text-input

addon/templates/components/frost-textarea.hbs
```diff
@@ -12,7 +13,14 @@
   tabindex=tabindex
 }}
 {{#if showClear}}
-  {{#frost-button class='clear' type='icon' tabIndex='-1' onClick=(action 'clear')}}
-    {{frost-icon  icon='close'}}
+  {{#frost-button
```
sglanzer-fire: Should just replace this whole thing with a div/svg target like we have in `frost-text`

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/153/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #158, opened 2016-08-02
fix bug when property is not defined

sandersky:
# PATCH
# CHANGELOG
- **Fixed** bug in `treeForAddon` that can affect consumers.

### DIFF
--- a/index.js
+++ b/index.js
@@ -156,7 +156,7 @@ module.exports = {
 
     const svgPath = path.join(this.project.root, 'public/svgs')
     if (isLegacy && fs.existsSync(svgPath)) {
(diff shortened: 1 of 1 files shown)


sglanzer-fire:
This should never be the case so I'm not sure why it would be failing - the frost pack comes from core and so does the icon pack code.  All this does is add legacy icons from public/svgs to that pack

sglanzer-fire:
This is only used to provide names to the demo anyway- might as well merge it

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/158/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #167, opened 2016-08-12
Remove most uses of lodash methods

sandersky:
# PATCH
# CHANGELOG
- **Removed** most uses of `lodash` methods in favor of ES6 and Ember methods.

### DIFF
--- a/addon/components/frost-checkbox.js
+++ b/addon/components/frost-checkbox.js
@@ -1,6 +1,5 @@
-import _ from 'lodash'
 import Ember from 'ember'
-const {Component, isEmpty, run} = Ember
+const {Component, isEmpty, run, typeOf} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
 import layout from '../templates/components/frost-checkbox'
@@ -101,7 +100,7 @@ export default Component.extend(PropTypeMixin, {
 
     input () {
       let id = this.get('value')
-      if (_.isFunction(this.attrs['onInput'])) {
+      if (typeOf(this.attrs['onInput']) === 'function') {
         this.attrs['onInput']({
           id: isEmpty(id) ? this.get('id') : id,
           value: this.$('input').prop('checked')
--- a/addon/components/frost-multi-select.js
+++ b/addon/components/frost-multi-select.js
@@ -1,4 +1,3 @@
-import _ from 'lodash'
 import computed, {readOnly} from 'ember-computed-decorators'
 import FrostSelect from './frost-select'
 import layout from '../templates/components/frost-multi-select'
@@ -58,9 +57,9 @@ export default FrostSelect.extend({
   },
 
   getValues () {
-    const selected = this.get('selectedItems')
+    const selected = this.get('selectedItems') || []
     const state = this.get('reduxStore').getState()
-    return _.map(selected, (item) => {
+    return selected.map((item) => {
       return state.baseItems[item].value
     })
   },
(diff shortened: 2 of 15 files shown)


addon/components/frost-multi-select.js
```diff
@@ -58,9 +57,9 @@ export default FrostSelect.extend({
   },
 
   getValues () {
-    const selected = this.get('selectedItems')
+    const selected = this.get('selectedItems') || []
```
sglanzer-fire, line 60: We may want to get in the habit of wrapping things like `[]` in `Ember.A([])` in case we (or someone else) has prototype extensions disabled - which would make operations like `map` unavailable to the vanilla `[]`

Just a note, nothing urgent

addon/components/frost-checkbox.js
```diff
@@ -101,7 +100,7 @@ export default Component.extend(PropTypeMixin, {
 
     input () {
       let id = this.get('value')
-      if (_.isFunction(this.attrs['onInput'])) {
+      if (typeOf(this.attrs['onInput']) === 'function') {
```
sglanzer-fire, line 103: We can start dropping `attrs` in these cases - `this.onInput` works equally well (and is nicer code)

addon/components/frost-select.js
```diff
@@ -145,8 +149,8 @@ export default Component.extend(PropTypeMixin, {
    */
   invalidFilter (data, displayItems) {
     return (
-      _.isArray(data) &&
-      _.isArray(displayItems) &&
+      Array.isArray(data) &&
```
sglanzer-fire, line 152: Should we be using destructing in this case as well?

`const { isArray } = Array`

tests/dummy/app/pods/icons/controller.js
```diff
@@ -15,10 +14,10 @@ export default Controller.extend({
   ],
   backgroundColor: 'bg-tile-color',
   iconPacks: computed('iconPacks', () => {
-    return _.map(_.keys(iconPacks), (name) => {
+    return Object.keys(iconPacks).map((name) => {
       return {
-        name: _.capitalize(name),
-        icons: _.map(iconPacks[name], (icon) => {
+        name: Ember.String.capitalize(name),
```
sglanzer-fire, line 19: Because I felt like figuring it out:

```
const { 
  String: { 
    capitalize 
  } 
} = Ember
```

tests/dummy/app/pods/icons/controller.js
```diff
@@ -15,10 +14,10 @@ export default Controller.extend({
   ],
   backgroundColor: 'bg-tile-color',
   iconPacks: computed('iconPacks', () => {
-    return _.map(_.keys(iconPacks), (name) => {
+    return Object.keys(iconPacks).map((name) => {
```
sglanzer-fire, line 17: Same comment as for Array - `const { keys } = Object`

sglanzer-fire:
Approved - just added a few notes on destructuring, but we can address that afterward

tests/dummy/app/pods/icons/controller.js
sandersky, line 19: 💪

addon/components/frost-select.js
sandersky, line 152: I see no reason not to.

addon/components/frost-multi-select.js
sandersky, line 60: But `[].map()` is valid ES6 and our code goes through Babel 😼

sglanzer-fire, line 60: Even better - ignore my junk 👍

PR #168, opened 2016-08-14
invoke child included hooks

### LINKED ISSUE #39: BLOCKER! Issue started from yesterday
loader.js:219 Uncaught Error: Could not find module `dummy/config/environment` imported from `ember-get-config/index`

I've started seeing this error after updating bower/npm. Seems like some upgrade spoiled the party

briarsweetbriar:
# patch

@sandersky 

This PR should resolve https://github.com/ciena-blueplanet/ember-prop-types/issues/39. Starting in the `0.1.0` branch, `ember-get-config` now uses the `included` hook, and as far as I know, that has to be manually invoked by the parent addon. That should have been included in my docs, and I'll be adding this step soon. Let me know if you know of another/preferred approach.


sandersky:
@sglanzer I would love to get your thoughts on this PR.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-frost-core",
-  "version": "0.22.2",
+  "version": "0.22.3",
```
sandersky: Dont' manually bump the version as the CI takes care of this when PR's are merged.

briarsweetbriar:
For reference, I found this approach [here](https://github.com/ember-cli/ember-cli/issues/3718).

sglanzer-fire:
Looking

sglanzer-fire:
@rwjblue endorsed - works for me - nested addon support is severely lacking for how mature the addon ecosystem should be at this stage.  We're going to need to update 20+ repos with this.

Thanks for the catch @null-null-null

sglanzer-fire:
Approved - just need to fix CI

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/168/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
I solved something like this before using a clunkier method where you check for this.app and if it doesn't exist look up the chain for the app, set it locally, then call include - this is better

sglanzer-fire:
Ah, just a linting error (we use a standard without semicolons)

index.js
```diff
@@ -81,6 +81,8 @@ module.exports = {
       this.app = app = app.app
     }
 
+    this.eachAddonInvoke('included', [app]);
```
sglanzer-fire: Drop the semicolon here to pass linting

briarsweetbriar:
Fix up! Let me know if you'd like me to squash.

briarsweetbriar:
> I solved something like this before using a clunkier method where you check for this.app and if it doesn't exist look up the chain for the app, set it locally, then call include - this is better

I think the `while (app.app)` hack is actually still necessary, but for different reasons. As I understand it, the argument that's passed into `included` is an [Ember CLI wrapper for your addon](https://ember-cli.com/api/classes/EmberAddon.html), which doesn't support `import`. So you've got to get its parent through `app.app`, which might be the actual app, or just another addon. Hence, the while loop. Once you have the parent app, you can finally call `import`.

I'm still pretty fuzzy on how all this works, though.

sglanzer-fire:
Approved - thanks again @null-null-null 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/168/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #176, opened 2016-08-24
Ensure addon works with Ember 2.1+

sandersky:
# PATCH
# CHANGELOG
- **Fixed** addon so it works with older versions of Ember all the way back to Ember `2.1`.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -7,6 +7,105 @@ module.exports = {
       }
     },
     {
+      name: 'ember-1-12',
+      bower: {
+        dependencies: {
+          'ember': '~1.12.0'
+        },
+        resolutions: {
+          'ember': '~1.12.0'
+        }
+      }
+    },
+    {
+      name: 'ember-1-13',
+      bower: {
+        dependencies: {
+          'ember': '~1.13.0'
+        },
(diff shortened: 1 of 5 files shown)


config/ember-try.js
```diff
@@ -7,6 +7,105 @@ module.exports = {
       }
     },
     {
+      name: 'ember-1-12',
```
rox163, line 10: you are just adding these props but the build is just tested against 2.1 right?

sandersky: Correct, I would have Travis test against all of them but due to Travis limiting us to 4 concurrent builds I'm not having CI run them all. By having them all defined here though a dev can easily run tests against a specific version via `ember try:one ember-2-3`.

sandersky: I should note this is the same config I'm using elsewhere, i.e. `ember-prop-types`. I'm just copy/pasting it into other addons.

rox163:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/176/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #184, opened 2016-08-26
Refining the frost-component blueprint

sglanzer-fire:
# fix
# CHANGELOG
- Updated the frost-component blueprint to reflect current best practices

### DIFF
--- a/blueprints/frost-component/files/__root__/__path__/__name__.js
+++ b/blueprints/frost-component/files/__root__/__path__/__name__.js
@@ -3,7 +3,7 @@ const { Component } = Ember
 import PropTypesMixin, { PropTypes } from 'ember-prop-types'
 import layout from '<%= templatePath %>'
 
-const <%= classifiedModuleName %> = Component.extend(PropTypesMixin, {
+export default Component.extend(PropTypesMixin, {
 
   // == Dependencies ==========================================================
 
@@ -30,5 +30,3 @@ const <%= classifiedModuleName %> = Component.extend(PropTypesMixin, {
   // == Actions ===============================================================
 
 })
-
-export default <%= classifiedModuleName %>
--- a/blueprints/frost-component/files/__root__/__path__/__name__.js
+++ b/blueprints/frost-component/files/__root__/__path__/__name__.js
@@ -1,13 +1,34 @@
 import Ember from 'ember'
-import FrostEvents from 'ember-frost-core/mixins/frost-events'
-import {
-  PropTypes
-} from 'ember-prop-types'
+const { Component } = Ember
+import PropTypesMixin, { PropTypes } from 'ember-prop-types'
+import layout from '<%= templatePath %>'
 
-const {
-  Component
-} = Ember
+const <%= classifiedModuleName %> = Component.extend(PropTypesMixin, {
 
-<%= importTemplate %>
-export default Component.extend(FrostEvents, {<%= contents %>
+  // == Dependencies ==========================================================
+
+  // == Component properties ==================================================
+
+  layout,
+
+  // == State properties ======================================================
+
+  propTypes: {
+    hook: PropTypes.string
+  },
+
+  getDefaultProps () {
+    return {
+      hook: '<%= dasherizedModuleName %>'
+    }
+  }
+
+  // == Functions =============================================================
+
+  // == Events ================================================================
+
+  // == Actions ===============================================================
+
 })
+
+export default <%= classifiedModuleName %>
--- a/blueprints/frost-component/index.js
+++ b/blueprints/frost-component/index.js
@@ -48,7 +48,10 @@ module.exports = {
     entityName = normalizeEntityName(entityName)
     return validComponentName(entityName)
   },
+
   locals: function (options) {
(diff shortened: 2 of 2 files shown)


sglanzer-fire:
Sample output:

```
import Ember from 'ember'
const { Component } = Ember
import PropTypesMixin, { PropTypes } from 'ember-prop-types'
import layout from '../templates/components/sample-component'

const SampleComponent = Component.extend(PropTypesMixin, {

  // == Dependencies ==========================================================

  // == Component properties ==================================================

  layout,

  // == State properties ======================================================

  propTypes: {
    hook: PropTypes.string
  },

  getDefaultProps () {
    return {
      hook: 'sample-component'
    }
  }

  // == Functions =============================================================

  // == Events ================================================================

  // == Actions ===============================================================

})

export default SampleComponent
```

blueprints/frost-component/files/__root__/__path__/__name__.js
```diff
@@ -1,13 +1,34 @@
...
+
+  // == Events ================================================================
+
+  // == Actions ===============================================================
 
-<%= importTemplate %>
-export default Component.extend(FrostEvents, {<%= contents %>
 })
+
+export default <%= classifiedModuleName %>
```
job13er: What's the reasoning for storing the component definition in a variable and then exporting it vs. just exporting the declaration?

blueprints/frost-component/files/__root__/__path__/__name__.js
```diff
@@ -1,13 +1,34 @@
...
-  PropTypes
-} from 'ember-prop-types'
+const { Component } = Ember
+import PropTypesMixin, { PropTypes } from 'ember-prop-types'
+import layout from '<%= templatePath %>'
 
-const {
-  Component
-} = Ember
+const <%= classifiedModuleName %> = Component.extend(PropTypesMixin, {
```
job13er: So, are we moving away from the initializer route then? Is it better to have the explicit mixin inclusion? I thought in previous discussions your position was that prop types was one of the scenarios where initializers were the preferred mechanism.

job13er: I really should use the word "route" when taking ember b/c clearly I didn't mean an ember route ;)

sophypal: I think @sandersky made the argument to keep it in because when you're unit testing, Ember initializers won't be run and tests will start to fail leaving the tester thoroughly confused if they never knew they were depending on `ember-prop-types`.

job13er: I think you mean @sandersky ;)

That sounds like a good reason to make it explicit all right :)

sophypal: Oh man, I just mentioned someone completely different. Corrected.

sandersky: Correct, if you rely on the initializer then `getDefaultProps()` isn't run in your integration tests and you start wondering why the hell props don't have their respective default values. I've spent hours around this and never found a way to easily auto-run an initializer for integration tests and quickly decided it isn't worth the pain.

sglanzer-fire: We've been running a few initializers using beforeEach and this could follow the same pattern, but it wouldn't be "automatic" - so until we figure out a solution to that it's best to include the mixin

blueprints/frost-component/files/__root__/__path__/__name__.js
sglanzer-fire: Positional params really https://guides.emberjs.com/v2.7.0/components/passing-properties-to-a-component/#toc_positional-params

job13er: Interesting. I'd be interested to see an example of when that's preferred.

sglanzer-fire: Helpers use it a fair amount - basically when the property you require is so intrinsically tied to the name of the component that it doesn't really make sense to "duplicate" it in a property.

Something like `{{play-sound 'foghorn'}}` instead of `{{play-sound sound='foghorn'}}`

job13er: OK, I can see that, for single properties. I doubt I'd ever see that two positional properties would be better than named properties though. Unless, of course, it's something like `add` and it doesn't matter which comes first. Seems like you'd know that ahead of time though, right? Like at the time you generate the component? Could it be an option in the blueprint? I'd hate to see all those pointless variable declarations just for the edge case of when you want positional properties.

sglanzer-fire: That's fair, I can move it out and just document how to do them in something like the twiddle/examples repo

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/184/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

Issue #179, opened 2016-08-25
Please update CSS in Frost select


baconC:
￼http://ciena-frost.github.io/ember-frost-core/#/select

.frost-select.open .drop-down-container ul {
-      height: auto;
-      box-shadow: 0 2px 6px 2px rgba(0,0,0,.2);
-      border-radius: 4px;
  }

.frost-select li {
-      border-bottom: solid 1px #F5F6F7;
-      color: #7b7f7f;
-      font-weight: 200;
-      height: 35px;
-      line-height: 35px;
-      padding: 0 10px;
-      width: 100%;
  }

.frost-select .up-arrow {
-      border-bottom: 12px solid #FFF;
-      border-left: 12px solid transparent;
-      border-right: 12px solid transparent;
-      height: 0;
-      width: 25px;
-      display: inline-block;
-      left: calc(50% - 15px);
-      overflow: visible;
-      pointer-events: none;
-      position: absolute;
-      right: 8px;
-      top: -12px;
-      -webkit-filter: drop-shadow(0 -2px 2px rgba(0,0,0,.15));
-      z-index: 2;
  }

.frost-select-list-item-selected {
    color: #231F20;
    font-weight: 500;
}

Issue #159, opened 2016-08-03
Frost Select throws error on enter without match


srowhani:
Hitting enter without selecting a value throws an error.
`Uncaught TypeError: Cannot read property 'value' of undefined`

``` js
  getValues () {
    const state = this.get('reduxStore').getState()
    return [state.baseItems[state.selectedItem].value]
  }
```

Should change to something like this:

``` js
  getValues () {
    const state = this.get('reduxStore').getState()
    const item = state.baseItems[state.selectedItem]
    return [item ? item.value : '']
  }

```

PR #185, opened 2016-08-27
Frost Select / Radio - Addressing several issues

srowhani:
# patch
# Changelog
- [x] There is a dead zone in the select between the text and the arrow. #183 
- [x] Please update CSS in Frost select #179
- [x] Frost Select throws error on enter without match #159
- [x] Using EventsProxy for events
- Frost Radio Group
  - Ember release 2.8 was erroring when destroying `parentView`
- Up arrow drawn in scss, one less element rendered
- Fix focus states by using attribute binding to `shouldOpen`
- If there are no display items, do not render `<div class='drop-down-container'>`

### DIFF
--- a/addon/components/frost-radio-button-input.js
+++ b/addon/components/frost-radio-button-input.js
@@ -1,13 +1,14 @@
 import Ember from 'ember'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
-import FrostEvents from '../mixins/frost-events'
 
 const {
   Component,
-  computed
+  computed: {
+    alias
+  }
 } = Ember
 
-export default Component.extend(FrostEvents, PropTypeMixin, {
+export default Component.extend(PropTypeMixin, {
   // == Properties =============================================================
   attributeBindings: [
     'checked',
@@ -15,15 +16,15 @@ export default Component.extend(FrostEvents, PropTypeMixin, {
     'value',
     'type'
   ],
-  classNames: ['frost-radio-button-input'],
-  excludeEvents: ['onChange'],
+  classNames: [
+    'frost-radio-button-input'
+  ],
   tagName: 'input',
 
   propTypes: {
     checked: PropTypes.bool,
     disabled: PropTypes.bool,
     hook: PropTypes.string,
-    value: PropTypes.bool,
     type: PropTypes.string
   },
 
@@ -35,17 +36,15 @@ export default Component.extend(FrostEvents, PropTypeMixin, {
   },
 
   // == Computed properties ====================================================
-  checked: computed('groupValue', 'value', function () {
-    return this.get('groupValue') === this.get('value')
-  }),
+
+  checked: alias('parentView.checked'),
 
   // == Functions ==============================================================
-  init () {
+  didInsertElement () {
     this._super(...arguments)
 
-    let assert = `${this.toString()} must be initialized in the yield block of 'frost-radio-button'`
+    let assert = `${this.toString()} must exist inside 'frost-radio-button'`
     let cond = /frost-radio-button/.test(this.parentView.toString())
     Ember.assert(assert, cond)
   }
-
 })
--- a/addon/components/frost-radio-button.js
+++ b/addon/components/frost-radio-button.js
@@ -1,13 +1,14 @@
 import Ember from 'ember'
 const {
   Component,
-  computed
+  get,
+  set,
+  typeOf
 } = Ember
-const {
-  readOnly
-} = computed
+
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
 import layout from '../templates/components/frost-radio-button'
+import computed from 'ember-computed-decorators'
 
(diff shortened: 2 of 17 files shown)


addon/components/frost-select.js
```diff
@@ -384,11 +384,17 @@ export default Component.extend(PropTypeMixin, {
   // ==========================================================================
   // Events
   // ==========================================================================
-
+  focusIn (event) {
+    this.get('reduxStore').dispatch(openDropDown)
+    // If an onFocus event handler is defined, call it
+    if (this.attrs.onFocus) {
+      this.attrs.onFocus(event)
```
sandersky: Didn't we deprecate usage of `onFocus` on the other core components? It seems we ought to do the same here.

srowhani: Good catch

sglanzer-fire:
Can you flag when this is ready for review again?

addon/components/frost-radio-button.js
```diff
@@ -39,57 +40,82 @@ export default Component.extend(PropTypeMixin, {
...
     Ember.assert(
       `${this.toString()} must be initialized with a 'value' property`,
-      this.get('value')
+      get(this, 'value')
     )
   },
-
+  // fix for Ember 2.8 on application destroy
+  willDestroyElement () {
+    set(this, 'parentView', null)
```
srowhani, line 103: `afterEach` hook from the `destroy-app` helper was throwing an error in 2.8. This addresses that.

sandersky: Care to elaborate?

addon/components/frost-radio-button-input.js
```diff
@@ -35,17 +36,15 @@ export default Component.extend(FrostEvents, PropTypeMixin, {
...
   // == Computed properties ====================================================
-  checked: computed('groupValue', 'value', function () {
-    return this.get('groupValue') === this.get('value')
-  }),
+
+  checked: alias('parentView.checked'),
 
   // == Functions ==============================================================
-  init () {
+  didInsertElement () {
```
juwara0, line 43: Why the change from init()?

addon/components/frost-radio-button.js
```diff
@@ -39,57 +40,82 @@ export default Component.extend(PropTypeMixin, {
...
-    target.id = this.get('groupId')
+    target.id = get(this, 'groupId')
     event.target = target
     return event
   },
 
   // == Events ===============================================================
 
-  init () {
+  didInsertElement () {
```
juwara0, line 91: Same question as above, why the change from init()?

addon/components/frost-radio-button.js
```diff
@@ -39,57 +40,82 @@ export default Component.extend(PropTypeMixin, {
...
-      if (this.get('disabled') || this.get('groupValue') === this.get('value')) {
+      if (get(this, 'disabled') || get(this, 'checked')) {
         return
       }
-      let change = this.get('onChange')
-      if (change && typeof change === 'function') {
-        change(this._createEvent(e, Ember.$(e.target).find('input')[0]))
+      const onChange = get(this, 'onChange')
+      if (typeOf(onChange) === 'function') {
+        onChange(this._createEvent(e, this.$(e.target).find('input')[0]))
```
juwara0, line 112: use `Ember.$` and destructure at top. I saw a few other cases where `Ember.$` is used that can be destructured as well.

addon/components/frost-radio-button-input.js
```diff
@@ -1,29 +1,30 @@
...
 const {
   Component,
-  computed
+  computed: {
+    alias
+  }
 } = Ember
 
-export default Component.extend(FrostEvents, PropTypeMixin, {
+export default Component.extend(PropTypeMixin, {
```
juwara0, line 11: Why is `FrostEvents` mixin being removed?

juwara0:
I would prefer to see this PR be focused on one area of changes instead of multiple changes over two different components. Keeping the PR focused makes the review easier to follow and if there are issues we can revert one set of changes without having to revert the entire PR with all of these changes.

juwara0:
@srowhani I am in the process of writing the test suites for frost-core components. I have already started looking at frost-radio-\* and have some of the changes you are making in that changeset. Ping me later and we can chat about it.

addon/components/frost-radio-button-input.js
sglanzer-fire: I believe this is an internal construct (radio buttons need to be part of a group), so my assumption is that we would have the events/proxy on the group interface

addon/components/frost-radio-button.js
sglanzer-fire: There is a difference between Ember.$ (global scope) and this.$ (component/local scope) in case that matters

sglanzer-fire:
Sorry to cause additional work @srowhani, but I agree with @juwara0 that it would be easier to merge this if the select/radio were separate PRs.  Would you mind splitting this?

PR #189, opened 2016-09-02
****FOR DISCUSSION**** Data driven button 

sglanzer-fire:
# minor

The intention would be to package this into a mixin that is compatible with prop-types so that we could effectively data drive any component simply by adding the mixin

This is the approach I'm exploring to expose all properties in situations like the modal cancel/confirm, but it will apply in a much wider variety of situations

### DIFF
--- a/addon/components/frost-button.js
+++ b/addon/components/frost-button.js
@@ -212,20 +212,35 @@ export default Component.extend(PropTypeMixin, {
 
   // == Events ================================================================
 
+  init () {
+    this._super(...arguments)
+
+    if (this.hash) {
+      Object.keys(this.hash).forEach((key) => {
+        const value = this.get(`hash.${key}`)
+        if (Ember.typeOf(value) === 'function') {
+          Ember.defineProperty(this, key, undefined, value)
+        } else {
+          Ember.defineProperty(this, key, Ember.computed.alias(`hash.${key}`))
+        }
+      })
+    }
+  },
+
   onclick: Ember.on('click', function (event) {
     if (!ViewUtils.isSimpleClick(event)) {
       return true
     }
 
-    if (!this.get('disabled') && typeOf(this.attrs.onClick) === 'function') {
-      this.attrs.onClick(this.get('id'))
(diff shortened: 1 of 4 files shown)


tests/dummy/app/pods/button/template.hbs
```diff
@@ -422,4 +422,23 @@
...
 
+  <div class='section-title'>Data driven</div>
+  <hr>
+  <div class='section'>
+    <div class='example'>
+      <div class='title'>hash</div>
+      <div class='demo'>
+        {{! BEGIN-SNIPPET button-hash }}
+        {{frost-button
+          hash=attrsHash
```
sglanzer-fire, line 433: This is the significant part of the interface

tests/dummy/app/pods/button/controller.js
```diff
@@ -1,9 +1,27 @@
...
   vertical: false,
 
+  // BEGIN-SNIPPET button-hash
+  createAttrsHash: Ember.on('init', function () {
+    this.set('attrsHash', {
+      priority: 'primary',
+      size: 'medium',
+      text: 'Data driven',
+      onClick: createActionClosure.call(this, this.actions.onClickHandler)
+    })
```
sglanzer-fire, line 21: Can absolutely clean up the DX here, but the point is that the button's attrs were defined outside of the template, can include anything the button supports and could easily be data driven

addon/components/frost-button.js
```diff
@@ -212,20 +212,35 @@ export default Component.extend(PropTypeMixin, {
 
   // == Events ================================================================
 
+  init () {
```
sglanzer-fire, line 215: Ordering between this and prop-types needs to be solved

sglanzer-fire: Confirmed in this twiddle https://ember-twiddle.com/516529ad06c87d185ae98c89e396ace1?openFiles=mixins.first.js%2Ctemplates.components.my-component.hbs that we can just order a proxy mixin like this to the right of the PropTypes mixin and it'll work fine

So this would be something like `Component.extend(PropTypesMixin, HashProxyMixin`

PR #191, opened 2016-09-21
Use onclick handler from parameter passed from component/controller property

nskrypnik:
# MINOR


psbanka (approved):
👍

sglanzer-fire:
What does this cover?  From the comment it appears to be something like

```
{{frost-button
  onClick=(action (mut foo) 'bar'))
}}
```

Can you confirm?

sglanzer-fire:
@nskrypnik @psbanka ^ can either of you clarify?

addon/components/frost-button.js
```diff
@@ -210,15 +210,28 @@ export default Component.extend(PropTypeMixin, {
...
   onclick: Ember.on('click', function (event) {
     if (!ViewUtils.isSimpleClick(event)) {
       return true
     }
 
-    if (!this.get('disabled') && typeOf(this.attrs.onClick) === 'function') {
-      this.attrs.onClick(this.get('id'))
+    const onClickHandler = this._getOnClickHandler()
+    if (onClickHandler && !this.get('disabled')) {
+      onClickHandler(this.get('id'))
```
sglanzer-fire, line 234: Why id?

PR #193, opened 2016-09-27
Fix a problem where frost-select would clear prompt on redraw

TheAxeMan301:
# PATCH

Fix a problem where frost-select would clear prompt on redraw

### DIFF
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -223,7 +223,7 @@ export default Component.extend(PropTypeMixin, {
     this._super(...arguments)
     const stateAttrs = this.get('stateAttributes')
 
-    const reduxAttrs = _.chain(stateAttrs)
+    let reduxAttrs = _.chain(stateAttrs)
     .map((attrName) => {
       const split = _.filter(attrName.split(ATTR_MAP_DELIM))
       let attrValue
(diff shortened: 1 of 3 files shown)


vesper2000:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/193/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-select.js
```diff
@@ -240,15 +240,18 @@ export default Component.extend(PropTypeMixin, {
     .fromPairs()
     .value()
 
-    this.get('reduxStore').dispatch(resetDropDown(reduxAttrs))
-
     const dataChanged = isAttrDifferent(newAttrs, oldAttrs, 'data')
     const selectedChanged = isAttrDifferent(newAttrs, oldAttrs, 'selected')
     const selectedValueChanged = isAttrDifferent(newAttrs, oldAttrs, 'selectedValue')
 
+    if (!(selectedChanged || selectedValueChanged)) {
```
sandersky, line 247: I find that `if (!selectedChanged && !selectedValueChanged) {` reads a bit better as it doesn't have the extra parenthesis that the negation applies to.

sglanzer-fire:
Makes sense, ideally we should get the test working - can we put another few hours of effort into that (maybe consult @juwara0) and if we can't get it in that time range we'll fall back to the manual test for now?

TheAxeMan301:
Added a test that does fail without my fix. Thanks @juwara0 !

sglanzer-fire:
Approved - thanks 👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/193/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #196, opened 2016-09-30
Fixed null selectedValue bug

TheOtherDude:
# PATCH

Couldn't decide if it was better to do the check in redux or in the keyUp handler (this is my first experience with redux). I chose redux because I guess it's supposed to handle state...will change it if requested, though!
# Changelog
- Pressing enter to select an item when there is no item hovered will no longer reset the selected value or cause an error
- Changing select data after a selection has been made will no longer cause an error.

### DIFF
--- a/addon/reducers/frost-select.js
+++ b/addon/reducers/frost-select.js
@@ -219,6 +219,9 @@ export default function reducer (state, action) {
       nextState = select(state, action.itemIndex)
       break
     case SELECT_HOVER:
+      if (state.hoveredItem === null) {
+        return state
+      }
       nextState = select(state, state.hoveredItem)
       break
     case CLICK_ARROW:
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -244,7 +244,7 @@ export default Component.extend(PropTypeMixin, {
     const selectedChanged = isAttrDifferent(newAttrs, oldAttrs, 'selected')
     const selectedValueChanged = isAttrDifferent(newAttrs, oldAttrs, 'selectedValue')
 
-    if (!(selectedChanged || selectedValueChanged)) {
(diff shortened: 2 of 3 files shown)


addon/reducers/frost-select.js
```diff
@@ -219,6 +219,9 @@ export default function reducer (state, action) {
       nextState = select(state, action.itemIndex)
       break
     case SELECT_HOVER:
+      if (state.hoveredItem === null) {
+        return state
+      }
```
agonza40, line 224: Do you mind making this section:

``` javascript
  if (state.hoveredItem === null) {
     nextState = state
  } else {
    nextState = select(state, state.hoveredItem)
  }
```

?
Having one return statement has been helpful during debugging.

TheOtherDude: The problem is that `nextState.lastAction` is set outside the switch statement, and `lastAction` is used to determine if onChange should be called, which gets the values to pass up in `getValues`, which is where the bug really is:

```
getValues () {
    const state = this.get('reduxStore').getState()
    return [state.baseItems[state.selectedItem].value]
  }
```

Do you want to just let the state progress as usual but rewrite that function to be:

```
  getValues () {
    const state = this.get('reduxStore').getState()
    const values = state.selectedItem !== null ? [state.baseItems[state.selectedItem].value] : null
    return values
  }
```

?

TheOtherDude:
Let's wait to merge this. I want to squeeze another fix in.

agonza40:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/196/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

Issue #120, opened 2016-06-15
On the ember-frost-color color palette demo page several of the colors only have 5 digits


cstolli:
In the printed hex values.  The whole group of Teal colors and a couple greens it looks like?  not a big deal, but we should fix.

PR #209, opened 2016-10-17
Fix for color palette demo  [Issue #120]

rox163:
# patch
# CHANGELOG
- Fixed issue #120

### DIFF
--- a/tests/dummy/app/pods/components/demo-palette/component.js
+++ b/tests/dummy/app/pods/components/demo-palette/component.js
@@ -13,7 +13,9 @@ function rgb2hex (rgb) {
 
   rgb = rgb.match(/^rgb\((\d+),\s*(\d+),\s*(\d+)\)$/)
 
-  return `#${hex(rgb[1])}${hex(rgb[2])}${hex(rgb[3])}`
+  return hex(rgb[1]) === '0'
+    ? `#00${hex(rgb[2])}${hex(rgb[3])}`
+    : `#${hex(rgb[1])}${hex(rgb[2])}${hex(rgb[3])}`
 }
 
 export default Component.extend({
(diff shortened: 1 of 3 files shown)


tests/dummy/config/environment.js
```diff
@@ -4,7 +4,7 @@ module.exports = function (environment) {
     modulePrefix: 'dummy',
     podModulePrefix: 'dummy/pods',
     environment: environment,
-    rootURL: '/',
+    baseURL: '/',
```
rox163, line 7: the LTS prep work in progress will make this update properly.

tests/dummy/app/pods/components/demo-palette/component.js
```diff
@@ -13,7 +13,9 @@ function rgb2hex (rgb) {
 
   rgb = rgb.match(/^rgb\((\d+),\s*(\d+),\s*(\d+)\)$/)
 
-  return `#${hex(rgb[1])}${hex(rgb[2])}${hex(rgb[3])}`
+  return hex(rgb[1]) === '0'
```
sandersky: Would we run the risk of `rgb[2]` or `rgb[3]` being `0` as well? Instead should it be something like:

``` js
const hex = rgb.slice(1)
  .map((value) => value === '0' ? '00' : `${hex(value)}`)
  .join('')

return `#${hex}`
```

rox163: Doh! yes! Just so happens none of our current colour palette has that.

sandersky:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/209/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #212, opened 2016-10-26
added inline support for frost-link

quincyle:
This is a #PATCH# change.
# CHANGELOG
- **Added** inline support for frost-link.

### DIFF
--- a/addon/templates/components/frost-link.hbs
+++ b/addon/templates/components/frost-link.hbs
@@ -1,26 +1,26 @@
 {{#if (eq design 'info-bar')}}
-  <div class='content '>
-    <div class=' text'>
-      <div class='svg'>{{frost-icon icon=icon}}</div>
-      {{yield}}
-      </div>
-  </div>
+    <div class='content '>
+        <div class=' text'>
+            <div class='svg'>{{frost-icon icon=icon}}</div>
+          {{#if linkTitle}}{{linkTitle}}{{else}}{{yield}}{{/if}}
+        </div>
+    </div>
 {{else if (eq design 'in-line')}}
-   <div class='content text'>
-      {{yield}}
+    <div class='content text'>
+      {{#if linkTitle}}{{linkTitle}}{{else}}{{yield}}{{/if}}
     </div>
 {{else if priority}}
   {{#if (eq priority 'primary')}}
-    <div class='content text'>
-      {{yield}}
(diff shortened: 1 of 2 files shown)


addon/templates/components/frost-link.hbs
```diff
@@ -1,26 +1,26 @@
...
-  <div class='content '>
-    <div class=' text'>
-      <div class='svg'>{{frost-icon icon=icon}}</div>
-      {{yield}}
-      </div>
-  </div>
+    <div class='content '>
+        <div class=' text'>
+            <div class='svg'>{{frost-icon icon=icon}}</div>
+          {{#if linkTitle}}{{linkTitle}}{{else}}{{yield}}{{/if}}
```
sandersky, line 5: Where did `linkTitle` come from? If this is a new optional property please add it to the links propTypes as well as documentation in the dummy app.

tests/dummy/app/pods/link/template.hbs
```diff
@@ -120,6 +120,25 @@
       </div>
     </div>
 
+      <div class="example">
+          <div class="title">in-line</div>
+          <div class="demo">
+            {{frost-link 'In-line' 'link.min'
```
sandersky, line 126: I'm having trouble understanding how these dummy app additions relate to the changes in the component.

quincyle:
This PR will be submitted against @juwara0 's fork.

addon/templates/components/frost-link.hbs
quincyle: This is coming from built in Ember `link-to` component.
https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-glimmer/lib/components/link-to.js#L795

tests/dummy/app/pods/link/template.hbs
quincyle: link-to natively support an in-line format with its own template. By default, when consumer is using an in-line format, first positional param will be set onto the `linkTitle` prop of the link-to component, and then being used in the template. Since frost-link overwrite the default `link-to` template, this thing no longer works.

addon/templates/components/frost-link.hbs
sandersky: I'd still like to see it documented in the dummy app. I had no idea this existed nor would I had I not looked at this PR.

PR #213, opened 2016-10-26
Remove width on frost-select input

ghost:
discussed with @sandersky about how this `width` property no longer seems necessary on the frost-select input element.  I verified that removing it does not adversely affect the demo in the dummy app.


sglanzer-fire:
What's the effective area to launch the select?  A click anywhere in the input?

ghost:
I think we should have a click register anywhere in the input.  I now see that removing the width property is not the solution, as it narrows the trigger width in the dummy demo.  So, maybe we want input.trigger width = 100%?

ghost:
@sglanzer:  sorry, this PR was probably premature.  I am closing because we will revisit the issue within our consuming app...we see css issues with this component, but it could be our fault.

sglanzer-fire:
@lirwin I was originally thinking 100% for the inner input, let me know what you find

PR #214, opened 2016-10-31
Rewrite frost-select from the ground up

sandersky:
#MINOR#

# CHANGELOG

* Rewrote `frost-select` from the ground up to make it keyboard friendly and fix all of the issues we have been having with it. However I kept the same component API so consuming apps should still just work. CSS selectors did change though which will break any overriding CSS and tests that depend on the old selectors.


sglanzer-fire:
I haven't looked across the code yet, but this is a stellar job - nailed the behaviour, reduced the dependency stack and we can keep our core components in-house, really excellent work!

sglanzer-fire:
If you have time, I'd like to see the selected items in the multiselect treated like 
image

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/214/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #227, opened 2016-11-08
DO NOT MERGE - Fixed sass linting issues + increased node version

dafortin:
* Fixed all the sass linting issues
* Increased the node version from >= 5 to >=6
* Fix `ember-cli-code-coverage` version to 3.5.0

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -8,7 +8,10 @@ options:
 
 files:
   include: '+(addon|app|tests)/**/styles/**/*.s+(a|c)ss'
-  ignore: '**/dummy/app/styles//**/*.s+(a|c)ss'
+  ignore:
+    - '**/dummy/app/styles//**/*.s+(a|c)ss'
+    - '**/_flex.scss'
+    - '**/_normalize.scss'
 
 # Documentation on all rules available here: https://github.com/sasstools/sass-lint/blob/master/docs/rules
 # 0 - disabled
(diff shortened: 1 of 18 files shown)


.sass-lint.yml
```diff
@@ -237,7 +240,7 @@ rules:
 
   # enforce how deeply a selector can be nested (defaults to 2)
   # https://github.com/sasstools/sass-lint/blob/master/docs/rules/nesting-depth.md
-  nesting-depth: 2
+  nesting-depth: 3
```
rox163: does this change work? I think the syntax is 

```
nesting-depth:
    - 2
    -
      max-depth: 3
```

because the first line represents whether its reported as an error [2] or warning [1] and the second line is where you change the depth.

dafortin: I think 2 is an error and 3 a warning. I just change this from error to warning because we have way too many ocurrence of this to fix it right now. It's a complex task to do.

.sass-lint.yml
```diff
@@ -8,7 +8,10 @@ options:
 
 files:
   include: '+(addon|app|tests)/**/styles/**/*.s+(a|c)ss'
-  ignore: '**/dummy/app/styles//**/*.s+(a|c)ss'
+  ignore:
+    - '**/dummy/app/styles//**/*.s+(a|c)ss'
+    - '**/_flex.scss'
```
rox163: _flex is something we wrote so I think we should try to lint it. Whats sass complaining about?

.sass-lint.yml
rox163:
```
# 0 - disabled
 # 1 - warning
 # 2 - error

```

Change the max-depth instead if you can for now. Put it to whatever largest depth it complains about. I did this for a number of repos because trying to redo the css is a big job.

.sass-lint.yml
dafortin: Humm let me give it another try. I will show you the end result to make sure i'm not breaking anything :)

.sass-lint.yml
dafortin: ok 👍

PR #258, opened 2016-11-09
Fixed select tabIndex after being disabled

TheOtherDude:

- [X] #patch# - backwards-compatible bug fix

# CHANGELOG

* Fixed the tabIndex on frost-select that are disabled and then enabled

### DIFF
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -171,6 +171,13 @@ export default Component.extend(PropTypeMixin, {
 
   @readOnly
   @computed('disabled', 'tabIndex')
+  /**
+   * Get appropriate tab index
+   * disabled state changes.
+   * @param {Boolean} disabled - whether or not input is disabled
+   * @param {Number} tabIndex - tab index
+   * @returns {Number} tab index
+   */
   computedTabIndex (disabled, tabIndex) {
     return disabled ? -1 : tabIndex
   },
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -43,7 +43,7 @@ export default Component.extend(PropTypeMixin, {
   // == Properties ============================================================
 
   attributeBindings: [
-    'tabIndex'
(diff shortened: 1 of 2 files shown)


addon/components/frost-select.js
```diff
@@ -170,6 +170,12 @@ export default Component.extend(PropTypeMixin, {
   },
 
   @readOnly
+  @computed('disabled', 'tabIndex')
+  computedTabIndex (disabled, tabIndex) {
+    return disabled ? -1 : tabIndex
+  },
```
sandersky, line 183:
``` js
  @readOnly
  @computed('disabled', 'tabIndex')
  /**
   * Get appropriate tab index
   * disabled state changes.
   * @param {Boolean} disabled - whether or not input is disabled
   * @param {Number} tabIndex - tab index
   * @returns {Number} tab index
   */
  computedTabIndex (disabled, tabIndex) {
    return disabled ? -1 : tabIndex
  },
```

tests/integration/components/frost-select-test.js
```diff
@@ -412,6 +412,45 @@ describeComponent(...integration('frost-select'), function () {
         expect(onChange.callCount, 'OnChange is not called').to.equal(0)
         expect(onFocus.callCount, 'onFocus is not called').to.equal(0)
       })
+
+      describe('when input is disabled and then re-enabled', function () {
+        beforeEach(function () {
+          this.set('disabled', true)
+          this.set('disabled', false)
```
sandersky: Doing two set calls back to back like this doesn't read well and doesn't guarantee they happen in different run loops. Instead can you do:

``` js
describe('when disabled', function () {
  beforeEach(function () {
    this.set('disabled', true)
  })

  describe('when re-enabled', function () {
    …
  })

  describe('when enabled and tab index changed', function () {
    …
  })
})

```

sandersky (changes requested):

sandersky (approved):

Issue #278, opened 2016-11-15
Write tests for frost-toggle _changeTarget()


juwara0:

juwara0:
Closed via: https://github.com/ciena-frost/ember-frost-core/pull/279

PR #279, opened 2016-11-15
relocated cloneEvent method, updated places where it is called and ad…

juwara0:
* **Added** unit tests for frost-toggle's _changeTarget()
* **Updated** location of cloneEvent()
* **Updated** frost-toggle.js and frost-radio-button.js to use new path for cloneEvent()

Closes: https://github.com/ciena-frost/ember-frost-core/issues/278

### DIFF
--- a/addon/utils/events.js
+++ b/addon/utils/events.js
@@ -1,8 +1,5 @@
 import Ember from 'ember'
-const {
-  $,
-  deprecate
-} = Ember
+const {deprecate} = Ember
 
 export default {
   addProperty: function (event, frostEvent) {
@@ -46,13 +43,6 @@ export default {
         }
       })
     })
-  },
-
-  cloneEvent: function (event, target) {
-    let newEvent = $.Event(null, event)
-    let newTarget = $.clone(target)
-    newEvent.target = newTarget
-    return newEvent
   },
 
   map: {
--- a/addon/utils/utils.js
+++ b/addon/utils/utils.js
@@ -1,0 +1,11 @@
+import Ember from 'ember'
+const {$} = Ember
+
+export function cloneEvent (event, target) {
+  let newEvent = $.Event(null, event)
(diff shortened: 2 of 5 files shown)


addon/utils/events.js
```diff
@@ -1,8 +1,5 @@
 import Ember from 'ember'
-const {
-  $,
-  deprecate
-} = Ember
+const {deprecate} = Ember
```
dafortin, line 2: What are you deprecating here?

addon/utils/utils.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember'
```
dafortin, line 1: Should we find a better name than `utils`? I feel like having a utils class in a utils folder is a bit utility overkill :P

addon/utils/utils.js
```diff
@@ -0,0 +1,11 @@
...
+const {$} = Ember
```
dafortin, line 2: Are you planning to add tests for this util?

addon/utils/events.js
juwara0: https://github.com/juwara0/ember-frost-core/blob/62cb57155987c6148c1b9ad44f82b6ee93c10b72/addon/utils/events.js#L33

dafortin: Oh it was missing. Got it :) 👍

addon/utils/utils.js
dafortin: I will add the tests in a future PR.

addon/utils/utils.js
dafortin: I think we can live with this after discussion.

addon/utils/utils.js
juwara0: @dafortin Agreed to write tests for this util method.

PR #289, opened 2016-11-17
Cleanup comment blocks

job13er:

# CHANGELOG

 * **Updated** existing component definitions to match recently updated generator.
 * **Added** a single export point at `addon/index.js`


addon/components/frost-button.js
```diff
@@ -214,8 +232,9 @@ export default Component.extend(PropTypeMixin, {
     }
   },
 
-  // == Events ================================================================
+  // == DOM Events ============================================================
```
sandersky, line 235: I thought it was `Events` because it could also include things like `didReceiveAttrs`, `init`, etc. Are we now splitting these into two sections?

sandersky (approved):

sandersky:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/289/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-button.js
job13er: Yeah, `DOM Events` and `Lifecycle Hooks`

job13er: First we replaced `Events` with `Lifecycle Hooks` then realized we actually _do_ use legit events, so we added `DOM Events`

dafortin:
Don't merge it please I will like to review it.

addon/components/frost-button.js
```diff
@@ -1,14 +1,11 @@
...
 import Ember from 'ember'
-const {
-  Component,
-  get,
-  isEmpty,
-  Logger,
-  typeOf,
-  ViewUtils
-} = Ember
+const {Component, Logger, ViewUtils, get, isEmpty, typeOf} = Ember
```
dafortin, line 5: Why are we moving to inline here? I thought we were keeping those on multiple lines.

addon/components/frost-button.js
```diff
@@ -69,11 +74,22 @@ export default Component.extend(PropTypeMixin, {
     text: PropTypes.string,
     title: PropTypes.string,
     type: PropTypes.string,
-    vertical: PropTypes.bool
+    vertical: PropTypes.bool,
+
+    // state
```
dafortin, line 79: Are we leaving the // even if there is nothing under that section? I know that we did delete a few empty comment blocks.

addon/components/frost-button.js
```diff
@@ -69,11 +74,22 @@ export default Component.extend(PropTypeMixin, {
...
+
+    // keywords
+    attributeBindings: PropTypes.arrayOf(PropTypes.string),
```
dafortin, line 82: I thought we were not putting any prop types on those?

addon/components/frost-button.js
```diff
@@ -85,6 +101,8 @@ export default Component.extend(PropTypeMixin, {
       title: null,
       type: 'button',
       vertical: false
+
+      // state
```
dafortin, line 105: same comment as above

addon/components/frost-button.js
```diff
@@ -227,12 +246,17 @@ export default Component.extend(PropTypeMixin, {
...
+  // FIXME: jsdoc
   _onFocus: Ember.on('focusIn', function (e) {
     // If an onFocus handler is defined, call it
     if (this.attrs.onFocus) {
       this.attrs.onFocus()
     }
-  })
+  }),
+
+  // == Lifecycle Hooks =======================================================
```
dafortin, line 257: Same comment remove empty blocks.

addon/components/frost-button.js
```diff
@@ -227,12 +246,17 @@ export default Component.extend(PropTypeMixin, {
...
 
   // == Actions ===============================================================
+
+  actions: {}
```
dafortin, line 261: Why an empty actions here?

dafortin:
I'm not sure to understand why we are changing all the components in this PR.

addon/components/frost-scroll.js
```diff
@@ -42,28 +53,28 @@ export default Component.extend(PropTypeMixin, {
   _setupPerfectScroll () {
     const debouncePeriod = 150
 
-    scheduleOnce('afterRender', this, () => {
+    run.scheduleOnce('afterRender', this, () => {
```
dafortin, line 56: Why removing the deconstructing here?

addon/components/frost-select-li.js
```diff
@@ -22,10 +35,16 @@ export default Component.extend(PropTypeMixin, {
     onSelect: PropTypes.func.isRequired
   },
 
+  /** @returns {Object} the default property values when not provided by consumer */
```
dafortin, line 38: Why adding an empty default?

addon/components/frost-toggle.js
```diff
@@ -78,42 +113,40 @@ export default Component.extend(PropTypeMixin, FrostEventsProxy, {
...
+  // == Actions ===============================================================
+
   actions: {
     /* eslint-disable complexity */
+    // FIXME: eslint
     _onClick (event) {
       if (get(this, 'disabled')) return
 
-      if (!isSimpleClick(event)) {
+      if (!ViewUtils.isSimpleClick(event)) {
```
dafortin, line 149: Deconstructing?

addon/components/frost-toggle.js
```diff
@@ -130,8 +163,8 @@ export default Component.extend(PropTypeMixin, FrostEventsProxy, {
       }
     },
     /* eslint-enable complexity */
-    noop () {
-      // we have this method to avoid action undefined error
-    }
+
+    /** A no-op action to avoid an action undefined error */
```
dafortin, line 167: Why doing comments line like this? I think it's clearer like this

```
/**
 * ....
 */
```

addon/index.js
```diff
@@ -0,0 +1,27 @@
+/**
```
dafortin, line 1: I don't understand the purpose of this. Maybe I'm missing something.

tests/unit/components/frost-radio-group-test.js
```diff
@@ -42,8 +42,8 @@ describeComponent(
 
       expect(
         component.get('inputs'),
-        'inputs: undefined'
-      ).to.be.undefined
+        'inputs: []'
+      ).to.be.eql([])
```
dafortin, line 46: we are not using the .to.have.length?

addon/templates/components/frost-button.hbs
```diff
@@ -1,3 +1,4 @@
+{{! Template for the frost-button component }}
```
dafortin, line 1: I don't understand why we are adding a comment like this in every component templates.

dafortin:
I got a few general comments:
* I would prefer to keep destructured items on mulitple lines when there is more than one listed
* I do not understand the addon/index.js exporting all
* I do not understand the need to create all the extra comments, blocks, etc. (// --DOM events----,etc...) I think we should at least remove them if they are empty. We recently did a big clean up of this. I understand the goal of adding all those sections, but I feel like there is a lot of sections and that it makes the code hard to read.
* I do not understand the need to have propTypes on Ember's classNameBindings, attributeBindings, classNames and layout properties
* I do not understand the need to create `actions: {}` when it is not needed.

dafortin (changes requested):
I got a few general comments:
- I would prefer to keep destructured items on mulitple lines when there is more than one listed
- I do not understand the addon/index.js exporting all
- I do not understand the need to create all the extra comments, blocks, etc. (// --DOM events----,etc...) I think we should at least remove them if they are empty. We recently did a big clean up of this. I understand the goal of adding all those sections, but I feel like there is a lot of sections and that it makes the code hard to read.
- I do not understand the need to have propTypes on Ember's classNameBindings, attributeBindings, classNames and layout properties
- I do not understand the need to create `actions: {}` when it is not needed.

dafortin:
I think we should discuss and determine the way to go on this one. If everyone is agreeing with those changes then we can move forward :) and it's nothing against anybody and nothing personnal. Those changes represent major changes in the core that we have been preparing for weeks :).

addon/components/frost-button.js
job13er: I've never heard it preferred to be multiple lines unless it doesn't fit on a single line.

addon/components/frost-button.js
job13er: That was the plan, but I can remove them  if it bothers you, I just figured, it would be much more likely that additional properties, when added would end up in the right place if the comment block for them was present. I noticed a few components which had events/functions mixed together probably b/c there were no events at one point and when they were added, the comment block wasn't added too.

addon/components/frost-button.js
job13er: It doesn't work to assign them in `getDefaultProps()` but I see no reason not to have the type checking still.

addon/components/frost-button.js
job13er: Same response as above.

addon/components/frost-button.js
job13er: Same response on why I don't think it's a good idea to remove empty blocks.

addon/components/frost-button.js
job13er: B/c it lets you always have a trailing comma on methods above it, instead of having to worry about having one or not depending on if it's currently the last one.

addon/components/frost-button.js
sandersky: :+1:

addon/components/frost-scroll.js
job13er: b/c nested deconstructing is evil. I can add it back, but I would not add it as it was, I'd add it as two deconstructing lines.

addon/components/frost-button.js
sandersky: I certainly prefer one line unless it gets too long.

addon/components/frost-select-li.js
job13er: b/c I was making the components look as they would when the generator runs. Again, for the same reason as the empty comment blocks, when changes are made, I think it makes sense to have the proper place to put things already there.

addon/components/frost-button.js
sandersky: Originally I was all for removing comment blocks when not using them but I've found it is more of a pain if I have to go back and add them, especially if it is the long ones like `DOM Events`.

dafortin: Do we have this documented? It might be easier for people if it's documented. I think we should discuss this before removing those. Again I'm not speaking for everyone :) I'm speaking for me only. So It probably worth it to have a discussion.

addon/components/frost-button.js
sandersky: I see no harm in doing so.

addon/components/frost-button.js
sandersky: That is a valid point I never thought of regarding the trailing comma.

addon/components/frost-button.js
dafortin: I personally prefer multiple lines for clarity. I think we need to establish a standard here, because I might be wrong.

addon/components/frost-button.js
dafortin: But it's kind of polluting the code.

addon/components/frost-toggle.js
job13er: Again, nested deconstructing makes my eyes bleed. But I can deconstruct it if you want. I don't believe we have any kind of rule that says everything must be deconstructed to the lowest possible level. The reason for deconstructing `Ember` was b/c they're soon gonna be doing named exports (or at least, that was my understanding). There is a downside when deconstructing as well, in that it is much more difficult, actually, normally impossible to stub the method in a test if you use it deconstructed.

If you want to stub this in a test, you can get a handle on `ViewUtils` in the test code and do something like

``` js
sandbox.stub(ViewUtils, 'isSimpleClick').returns(true)
```

Whereas, if it's deconstructed at the top of the module, replacing the `isSimpleClick` on `VIewUtils` from the test module won't do you any good b/c the module under test already has a handle on the non-stubbed version and will call that instead.

addon/components/frost-button.js
dafortin: From what I understood @rox163 was wondering if we were agreeing on this. So maybe we should discuss this too.

addon/components/frost-toggle.js
job13er: I respectfully disagree on single line comments. But if we have a majority ruling that even single line comments should have the header and footer, I'll change it.

addon/index.js
job13er: The purpose of the `index.js` file in general? It allows someone to do something like:

``` js

import {Button} from 'ember-frost-core'
```

instead of 

``` js
import Button from 'ember-frost-core/components/frost-button'
```

By allowing the former, you don't consider the path to a component within your addon as part of your public API. If you export the thing at the root of the addon, then you can move it around within your addon (maybe someone decides pods are the way to go, or something like that) you don't impact your consumers at all. 

We do this in bunsen as well: https://github.com/ciena-frost/ember-frost-bunsen/blob/master/addon/index.js

job13er: Having a single export module at a`addon/index.js` is not unheard of, even `ember-computed-decorators` does it: https://github.com/rwjblue/ember-computed-decorators/blob/master/addon/index.js

tests/unit/components/frost-radio-group-test.js
job13er: Yes, previously the default was `undefined` I thought it safer for an array type to have a default of `[]`. I'm not sure I understand your question/concern.

dafortin: ok i can live with that one :) no big deal :)

addon/components/frost-button.js
rox163: I thought something about them didnt work, so now its clear that just the default assignment doesnt work. Im ok with type checking

addon/components/frost-button.js
juwara0: I prefer multiple lines because:
1. It is easy to be consistent on when to break the line. When is it too long? If there is more than x split onto a new line.
2. If I am working on something and I want to change what is destructured I can just comment out that line.
3. To me it is easier to read.

addon/templates/components/frost-button.hbs
job13er: It was in the generator and while it doesn't make as much sense here, when components are in pods, and they're all named `template.hbs` it's very handy to know what template you're looking at. I don't see how it hurts anything.

addon/index.js
dafortin: Interesting :)

addon/components/frost-button.js
dafortin: 👍

addon/components/frost-button.js
dafortin: We will have to add this in every component. I feel like we are overkilling it.

job13er:
@dafortin let me address these, although I already did with many of them in-line. 

> I would prefer to keep destructured items on mulitple lines when there is more than one listed
> I **strongly** disagree. And I know I'm not the only one. 
> 
> I do not understand the addon/index.js exporting all

I explained in detail on the file comment. It's to de-couple your addon's directory structure from your consumers. 

> I do not understand the need to create all the extra comments, blocks, etc. (// --DOM events---,etc...) I think we should at least remove them if they are empty. We recently did a big clean up of this. I understand the goal of adding all those sections, but I feel like there is a lot of sections and that it makes the code hard to read.

I can understand where you're coming from, but I believe the idea behind those sections was to have a place to put things for when the component changes. I am very confident that without them, when people add code that _should_ go into one of those sections, if the section isn't there and empty, they will not thing to add that section at the time they add the new piece of code.

I disagree that it makes things hard to read, it's just a couple comment lines. 

> I do not understand the need to have propTypes on Ember's classNameBindings, attributeBindings, classNames and layout properties

Probably not a need, just a want to be able to have the same type-checking on keyword properties as we have on others. We won't use the `getDefaultProps()` since it breaks some stuff, but I see no reason not to take advantage of the type checking, even if it's just in console warnings. 

> I do not understand the need to create actions: {} when it is not needed.

Again, it's not a need, it's just a convenience. I find it annoying to have to remember to delete the comma of the last function in my component when I don't have any actions. By including an empty actions object, you allow every method/property above to be treated the same, they each get a trailing comma.

addon/components/frost-toggle.js
rox163: Single line comments should stay consistently `//`
Imho Function description comments  above the function [or really long comments] should have header+footer.

job13er:
> I think we should discuss and determine the way to go on this one. If everyone is agreeing with those changes then we can move forward :) and it's nothing against anybody and nothing personnal. Those changes represent major changes in the core that we have been preparing for weeks :).

I'm a little confused. I thought we had already discussed most of this in slack and when I wrote the generators. All I did was make the components here match the generator here. 

Are you saying that I'm undoing major changes in the core that you've been working on for weeks? I find that hard to believe, all I did was add some comment blocks and move some code around.

addon/templates/components/frost-button.hbs
rox163: same. seems unnecessary.

addon/components/frost-toggle.js
job13er: JSDoc doesn't pick up `//`

job13er: OK, I was  trying to be nice and add the missing JSDoc comment, I guess I should have just gone with

``` js
// FIXME: jsdoc
```

On previous teams, I've had people complain about the verbosity of 

``` js
/**
 *
 */
```

for single-line function descriptions. I guess I just can't win.

dafortin:
> Are you saying that I'm undoing major changes in the core that you've been working on for weeks? I find that hard to believe, all I did was add some comment blocks and move some code around.

I was talking more about consistency. For sure you didn't break anything. I'm just not sure we all have the same standards or I might be missing some standards in my toolbox. I'm not aware of most of those.

addon/components/frost-button.js
job13er: 1. It's easy to be consistent both way, you switch to one per line as soon as it's too large to fit 120 column (our defined max-line-length)
2. I've never had a case when I wanted to do that. I'd like to understand that use-case more. 
3. I disagree 100%, I find it physically painful to look at the multi-line when it would have fit on one line. Similar to something like:

``` js
if ((foo === 'bar') && (bar === 'bad'}) {
  // do something
}
```

vs.

``` js
if ((foo === 'bar') &&
    (bar === 'baz')) {
  // do something
}
```

sandersky:
:+1:

addon/components/frost-button.js
dafortin: I understand you point. Those arguments are very valuable.

addon/components/frost-button.js
dafortin: Yeah you are probably right.

addon/components/frost-button.js
dafortin: it's ok it probably doesn't hurt.

addon/components/frost-button.js
dafortin: 👍

addon/components/frost-button.js
dafortin: 👍

addon/components/frost-scroll.js
dafortin: I understand you point

sglanzer-fire:
Gonna try to walk in here and diffuse anything not already covered in Slack discussions.  Give me an hour or two while I put the baby to bed.

dafortin:
I think your arguments are good and that it make sense :) We can rediscuss the details eventually if necessary.

dafortin (approved):

tests/unit/components/frost-radio-group-test.js
job13er: Sorry,  I totally misread that. I'm not sure why you'd favor `to.have.length()` over actually making sure it's the right thing. I guess when it's empty it's not a big deal, but when you expect something there, it'd be much preferred to use `.eql()` I'd think.

addon/components/frost-button.js
job13er: It is a bit annoying to have them duplicated though, maybe we can get `prop-types` to include them, or at least provide something we can add that let's us avoid the duplication, while still getting the type-checking.

addon/components/frost-button.js
juwara0: @job13er - Per my #2 comment - I did not communicate that point correctly. A better example would be adding or deleting an item. When the items are broken onto multiple lines, I find it quicker to just hit return and add the new item or conversely just use my keyboard shortcuts to highlight the line and hit delete.

I recognize that this is just a matter of preference but I wanted to clarify that and answer your question.

addon/components/frost-button.js
dafortin: We can if you want

addon/components/frost-button.js
sglanzer-fire: I believe we agreed in Slack before that single line was fine with a reasonable wrap - purely the judgement call of the person writing the file imo.

There are cases with nested destructuring where a single line doesn't look as good in my opinion, e.g.

```
const {
  Component,
  computed: {
    readOnly
  },
  run
} = Ember
```

vs
`const { Component, computed : { readOnly },  run } = Ember`

But I do believe this is a matter of preference and that we can have a few heuristics to guide these styles.

addon/components/frost-button.js
sglanzer-fire: I was originally asking to remove empty sections, but as long as the number of sections remains reasonably low, I'm fine with keeping them.

I want to avoid a situation where I have to scroll through reams of empty comments to see meaningful code (I've been in cases before with 80+ lines of empty comments per file).  I think we're avoiding that with the single line section headers and the low number of sections we have.

addon/components/frost-button.js
sglanzer-fire: Ha - I was just going to type "this seems a bit boilerplate/redudant" then saw @job13er's last comment, yes - let's do that

addon/components/frost-button.js
```diff
@@ -227,12 +246,17 @@ export default Component.extend(PropTypeMixin, {
     }
   }),
 
+  // FIXME: jsdoc
```
sglanzer-fire, line 249: I'm not a fan of putting jsdocs on _every_ function, if code is clear enough to be self-documenting then I feel the inclusion of jsdocs actually detracts from readability

addon/components/frost-button.js
sglanzer-fire: Is this in the new blueprint - I like the idea - the actions is going to end up as an object anyway and it gives new devs a really obvious hook point

addon/components/frost-checkbox.js
```diff
@@ -1,34 +1,50 @@
...
+  classNames: ['frost-checkbox'],
   layout,
 
+  // == PropTypes =============================================================
+
+  /**
+   * Properties for this component. Options are expected to be (potentially)
+   * passed in to the component. State properties are *not* expected to be
+   * passed in/overwritten.
+   */
```
sglanzer-fire, line 25: Do we need to explain this on every file?

addon/components/frost-checkbox.js
```diff
@@ -1,34 +1,50 @@
+/**
+ * Component definition for the frost-checkbox component
+ */
```
sglanzer-fire, line 3: This seems redundant - I think we need to make some basic assumptions that people know how Ember projects are structured

addon/components/frost-checkbox.js
```diff
@@ -76,18 +67,59 @@ export default Component.extend(PropTypeMixin, {
...
+      }
+      e.preventDefault()
+      e.stopPropagation()
+      return false
+    }
+  },
+
+  // == Lifecycle Hooks =======================================================
+
+  /* Ember.Component method */
```
sglanzer-fire, line 102: I commented on a similar case above, but this is a good example of what I'm talking about - if it's unclear that everything inside of Lifecycle Hooks is from the component, then we might as well call the header `Ember Component Lifecycle Hooks` and do away with the (in my opinion) redundant jsdocs

addon/components/frost-scroll.js
sglanzer-fire: Why is nested deconstruction evil?  I personally use this a fair amount

addon/components/frost-toggle.js
sglanzer-fire: If a jsdoc comment is necessary, I'd prefer the single line `/** */` format so that I don't have to look across too many meaningless lines when scanning a file

sglanzer-fire: In the case of a line-wrapped comment, absolutely go ahead an include a header/footer - I guess in general I would just be a fan of less comments, more readable code

addon/index.js
sglanzer-fire: We had agreed on this a week or so back, it's my fault for not recording it as an issue against the project

addon/templates/components/frost-button.hbs
sglanzer-fire: I'd prefer to rely on editors to handle distinguishing file names, Atom / VSC / WebStorm / Sublime are all capable of handling this

sglanzer-fire: It's fine for the generator to include this in the same way that it'll include usage comments (like mirage does).  I just delete them as soon as I move into a file because once you start writing it you don't really need the comments.

sglanzer-fire (approved):
There is a good amount of value here, let's get it merged and we'll adjust after the fact.  I do wish you would have asked in a few cases before changing all the files since not all of these are established conventions and a few conflict with the way we've decided to approach those yet-to-be-agreed-on areas, but your intentions were good.

sglanzer-fire:
Approved - we'll discuss and adjust some of this after the fact and establish more conventions while we do so.  I'll take point on this

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/289/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-checkbox.js
sglanzer-fire: Actually, it's fine to have this from the generator, I just wouldn't expect it to be kept afterward

addon/components/frost-scroll.js
sglanzer-fire: Saw your comment on the test stubs below - I can buy it, would still prefer to destructuring lines in this case

sglanzer-fire:
Raised #291 - #300 to track

addon/components/frost-button.js
job13er: I've never been a fan of nested destructuring. I don't think it fits with the reason why we were doing the destructuring of `Ember` to begin with. If I'm not mistaken, we were doing it b/c we expected ember to eventually move to named imports and 

``` js
import Ember from 'ember'
const {Components, computed, run} = Ember
```

could easily be converted to

``` js
import {Component, computed, run} from 'ember'
```

I don't believe we'd want

``` js
import Ember from 'ember'
const {
  Component,
  computed: {
    readOnly
  },
  run
} = Ember
```

to become 

``` js
import {
  Component,
  computed: {
    readOnly
  },
  run
}  from 'ember'
```

At least, I've never seen that kind of destructuring on an `import` (if it's even supported) and I'd much rather see

``` js
import {Component, computed, run} = Ember
const {readOnly} = computed
```

That's why, if we want to destructure one of the children of `Ember`, which has it's own downsides of doing (like making it harder to stub it in tests), I propose we do it like so:

``` js
const {Component, computed, run} = Ember
const {readOnly} = computed
```

job13er: I have a  hunch that you can't actually use destructuring inside an `import`, based on [this](https://esdiscuss.org/topic/import-statements-destructuring) which makes sense, b/c the import syntax is static and evaluated at compile time, not run time. Similar to why you can't use a variable name in your `from`.

addon/components/frost-button.js
job13er: Yes it is.

addon/components/frost-scroll.js
job13er: OK, I suppose it's not evil, although I did find it kinda ugly and hard to follow, but I don't believe it's in line with the reason why we destructure (I keep calling it deconstructing) the `Ember` object. At least, to the best of my memory, we were doing it b/c we wanted to mimic (and make it easy to switch to) named exports. You can't destructure a named export, so I don't think we should use nested destructuring of `Ember`. You can use nested destructuring of one of the named exports after the import line, so I wouldn't be opposed to use having a second line that uses nested destructuring if we want to go that route, but the main reason I didn't like the nested on the main `Ember` line was b/c it makes it so that now, it's a lot more work to convert to named exports, probably not automatable like it would be if we stuck with first destructuring the top-level attributes in one statement, then any other ones in follow-up statements.

addon/components/frost-toggle.js
job13er: So, when do we want to destructure vs not? I don't believe we have something like `rewire` that lets you swap out top-level functions for stubs within the module under test, so I think if there's any chance you'd want to stub the thing in a test, it's a better bet to not destructure it. That being said, I'm not sure if we ever want to stub something like `Ember.run.later` or `Ember.run.scheduleOnce` so I guess those are safe to destructure (not this comment, I realize, but I think `isSimpleClick` falls into the other category of something we _would_ want to stub, since it might be harder in the test to generate click events that are or aren't simple.

addon/components/frost-toggle.js
job13er: I'm not a big fan of big comment blocks either, but JSDoc on a function is my exception. I think it's far easier to enforce (in review at least) that all functions have JSDoc vs. trying to figure out when something is simple enough to not need it. For the simple ones, I like the `/** */` one liners for the same reason as @sglanzer. I'm not sure that multi-line paragraph comments are ever needed though. I guess the exception might be when discussing the top-level properties of a component. If it's talking about a block of code, I think the right move is to factor that block into a function and have that big comment block be the comment of the function. Comment blocks on chunks of code don't change with the chunk of code as much as comment blocks on functions change with the function. So you end up with less chance of stale comments (which are often worse than no comments) in that case.

addon/templates/components/frost-button.hbs
job13er: So are you saying leave it in if it's in a pod, b/c the file name is the same, but delete it if the file name is unique? I can live with that, I guess. I just like having header comments on files as it gives me a place to explain what's going on. Sometimes it's handy to be reminded of what the `tagName` is when looking at the template (info you don't have in that context). I guess that's more of a case-by case decision.

Are you suggesting we not allow people to have header comments on templates, or just suggesting that having them with no more information than the name of the component is pointless?

addon/components/frost-button.js
job13er: While in principle I agree with you, in practice I don't. I've found it's much more difficult to enforce "put JSDoc comments on appropriate functions" vs. "put JSDoc comments on all functions" If the function is simple, the doc can be one-line, if it has parameters, they should be documented though. 

It's also very easy for a function that was simple, to suddenly become less so. By having the jsdoc comment there, we have lint rules that ensure it's accurate (all params are covered, return type is covered, etc). If it doesn't have anything, the function can creep in complexity w/o being documented.

addon/components/frost-checkbox.js
job13er: I thought, since it's not common knowledge about the separation of 'options' and 'state' that it would be beneficial. If you think everyone already knows about that (which would surprise me, but it's been known to happen) I'm fine with removing it.

addon/components/frost-checkbox.js
job13er: Again, like the template header, it's mainly there b/c I find it horribly inconvenient to have every component definition named `component.js` and it's not at all obvious what component I'm working on, especially when I have multiple open and there's not room in the tab for the folder name and the file name. Really, these header blocks should include more information about the component, why it exists, what it's intended uses are, etc. But I was just filling in what the generator would put in there by default. I guess with simple things like checkbox there's not much  to describe, but even then it might be useful to know if it maintains it's own state, does the consumer just get change events and is responsible for reporting the updated state back to the component? Things of that nature. If you don't think that's a good idea, we can remove it, I just find modules w/o a header block describing what they are and what they're doing feel "naked" to me. I certainly could be in the minority there.

addon/components/frost-checkbox.js
job13er: First, it's not JSDoc `/*` vs `/**` if it _were_ JSDdoc, it'd fail lint, b/c we don't discuss params (well, I'm not sure on this particular one b/c I'm too lazy to look at the file view to see which function this is talking about, but you get the idea). The original reason for those comments was b/c our previous team wanted all functions to have _some_ header, b/c it's easy to see in review that a function is missing a header and complain than to have to figure out if this particular function should have a header or not. It was particularly useful when there weren't sections w/i the file delimited by comment blocks, or when more than one type of function was allowed within the same block. I agree that now that we have a section dedicated to these methods, it's less useful. Something about seeing an undocumented function still triggers my PR review senses though... probably b/c when doing a review of a new function being added, you don't get the context of what block it's in. I suppose I can get over that if everyone feels it's harmful to have them.

job13er:
> I do wish you would have asked in a few cases before changing all the files since not all of these are established conventions and a few conflict with the way we've decided to approach those yet-to-be-agreed-on areas, but your intentions were good.

I'm curious, who's the "we" in that statement, because I felt I was undoing exactly what you feel I was doing. From my point of view, these files had started to use conventions that were in conflict with how we (clearly a different "we") had decided to approach those yet-to-be-agreed-on areas. 

From the other addons/apps I've been working on, and I admit I haven't been in all of them, the changes that I made that were not in relation to just organizing things in the comment blocks, were moving this addon more in line with the others, as this one had started to use conventions that weren't agreed upon (at least not by anyone I talked with) and I didn't want people looking here and duplicating it elsewhere, since just about everyone will start looking here when beginning development. 

Clearly, we have some conflicting points of view here, but I'm not entirely sure the best way to resolve that.

addon/components/frost-button.js
job13er: @juwara0 Thanks for the explanation, that makes more sense to me than commenting one out temporarily did. I've actually liked doing things one-per-line like that in the past for exactly that same reason, it makes the diffs look nicer too. However, since we're not allowing trailing commas on everything, and changing which one has that trailing comma mitigates some of the benefit of the one-per-line approach, I've gotten used to the fact that I'm gonna have to edit lists of things that are in-line and not broken out. 

I don't think I'd want all my function parameters, for instance one per line, or all the dependencies of my computed property listed out one per line. I think that by breaking out the named imports one per line all the time, we're being inconsistent in that regard. If we're gonna do that, for all the reasons you mentioned, we should probably do it across the board, don't you think, with all lists of things?

Or do you think imports are separate enough from the others that the inconsistency isn't a big deal? I guess I could see that, they are the only bit that's processed statically at compile-time.

sglanzer-fire: I'm fine with the

```
const {Component, computed, run} = Ember
const {readOnly} = computed
```

style

@juwara0 can we align on an approach for single versus multi-line?  If so, I'll put this out to #frost-design as a new convention and close the associated issue

addon/components/frost-button.js
sglanzer-fire: Ok, so the proposal here will be to keep the sections (under the assumption they remain at a low number - which seems reasonable).  Once again, I'll put this out to #frost-design for any comments, then close the associated issue.

addon/components/frost-button.js
sglanzer-fire: I've found that enforcing something like this tends to shut off people's brains and I'd much rather have people be conscious of how their code is going to be read.  We're likely not going to come to an common agreement on this one, but here is my shot at it:
1.  When everything has comments, there is too much noise in a file - code gets harder to read and important comments can be missed since they don't contrast from the rest of the file.  
2.  Even simple comments require maintenance or they get out of sync with the source and I've seen enough of this in my past to be wary.  
3.  Many comments are redundant and effectively feel like writing double the code (I've seen debates over a single extra character).  
4.  People tend to downplay the need to name variables/functions appropriately believing comments cover all sins.  
5.  It's yet another syntax for people new to the domain to learn - necessary for some cases yes, but easier to lookup in those cases than learn by heart.

addon/components/frost-checkbox.js
sglanzer-fire: No, I'm ok with that and it's a good explanation for why you find there to be value.  Things like this in the header don't bother me nearly as much as inline comments because once I scroll past it there isn't any clutter in the file.   I'm good with keeping this - same drill I'll post this for comments in #frost-design, record the convention and close the issue.

addon/components/frost-checkbox.js
sglanzer-fire: I'm cool with keeping it in the generator, it falls in a class of "usage" comments that I would tend to delete - something like https://github.com/samselikoff/ember-cli-mirage/blob/master/blueprints/ember-cli-mirage/files/__root__/scenarios/default.js#L3 these.  This is kind of in the same vein as my view on inline comments above, so we can let that result dictate this one if that sounds reasonable.

addon/components/frost-checkbox.js
sglanzer-fire: I'm sure we're not debating wether or not this is jsdoc (you're right) - the point in this case would be that the section header speaks for the elements within it.  If we're going with inline comments then I would expect a description of what is being done in the hook moreso than a description of the hook.  Another one for the inline docs on all functions debate.

addon/components/frost-scroll.js
sglanzer-fire: Swapped my stance above on this - I'm fine with your approach

addon/components/frost-toggle.js
sglanzer-fire: We'll stick with single level

addon/templates/components/frost-button.hbs
sglanzer-fire: I'm fine with keeping these comment blocks in the header of files - no real harm and there is some obvious benefit for certain users.

addon/components/frost-button.js
sandersky: I'm in agreement to use

``` js
const {Component, computed, run} = Ember
const {readOnly} = computed
```

over

``` js
import {
  Component,
  computed: {
    readOnly
  },
  run
}  from 'ember'
```

as I find the latter much more eye bleeding. This is of course a personal preference.

sglanzer-fire:
"We" would be the developers on the UI platform team in this case, while we're not exclusively responsible for `core` it has traditionally been a domain we take care of - obviously "we" the community need more communication when there aren't established conventions, but it looks like we'll be able to clear up another set based on the discussion here.  Most of those hadn't come up in public discourse before to my knowledge (retained file headers, multi-destructuring, required jsdocs on functions).

I believe we have aligned proposals to put out to #frost-design for all of these discussions at this point minus inline jsdocs.  Does that align with your view?

addon/components/frost-button.js
juwara0: @job13er - You are welcome. I think the imports/destructing are separate enough that if we wanted to break them out it would be fine. However, I do see your point about the trailing commas. I am torn on this one. I see tradeoffs with each approach. In this case, I recommend that we just pick one and move forward.

addon/components/frost-button.js
job13er: I'm not sure I understand your proposal. 
1. Are you saying that not all methods should have JSDoc comments? Which ones should and which ones shouldn't? Is it a judgement call? If so, whose judgement? I've found myself often times looking to work on a function that someone else wrote and not understanding what the parameters are supposed to represent, it takes me quite a while to track down what they were trying to do, particularly if that person isn't available to discuss it. 
2. That is precisely why we have linting rules to make sure that additional parameters are reflected in the comment. If you're talking about the description of the method, then yes, that is something that I care a lot less about than the description of the inputs and outputs. (`@param` and `@returns`). 
3. I'm almost never in favor of writing more. I'm lazy, as all engineers should be. Lazy people automate things. The way we've addressed this in the past is by signing off that an obvious method doesn't require a "proper" comment, but `/* obvious */` can do. That distinguishes between someone consciously deciding to flag a method as obvious vs. just forgetting to add documentation to the method. Then in review, the merits of whether the method should qualify for `/* obvious */` treatment can be discussed on a case-by-case basis. 
4. I'm not sure how having comments gets anyone off the hook for naming a variable properly. That comes down to code review though. I don't think there's any automated system that can do that for us, and I don't think documenting a variable lowers the need for naming it properly to begin with. As long as we are in agreement to that statement, then I don't think documenting the variable hurts. People still need to review the code and sign off on it. Whether there are comments or not, that check still needs to happen. 
5. I don't buy this one. If someone is new to any domain, there are things to learn. I'd rank learning how to format 

``` js
/**
 * Description
 * @param {type} name - description
 * @returns {type} description
 */
```

below the bottom of anything else I've had to learn when writing javascript, particularly Ember. Plus, we have lint rules that tell you when it's not working. True, they can probably have better error messages, but saying that it's too much of a strain to have developers learn JSDoc is not what I'd consider a valid argument against it.

job13er:
Yeah

PR #329, opened 2016-11-30
Stop invoking child include() methods

job13er:

### Background

https://github.com/ciena-frost/ember-frost-core/pull/168 added this change, but according to https://github.com/null-null-null/ember-get-config/issues/10 it's no longer necessary as of `ember-get-config@0.1.8`. As it turns out, `ember-prop-types@3.0` (which we're currently using) is using `ember-get-config@0.1.11` (at the time of this PR) and so we should be safe to remove this. Plus, as of `ember-get-config@0.1.8` there's a depreciation notice when it's called again.  

# CHANGELOG

 * **Undo** the change from [#168](https://github.com/ciena-frost/ember-frost-core/pull/168) as it has been [rendered unnecessary](https://github.com/null-null-null/ember-get-config/issues/10)

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,8 +81,6 @@ module.exports = {
       this.app = app = app.app
     }
 
-    this.eachAddonInvoke('included', [app])
-
     this._super.included(app)
 
     if (app) {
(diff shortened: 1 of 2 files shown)


sglanzer-fire:
Approved 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/329/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

Issue #333, opened 2016-12-01
Spread: Links first 2 parameters (text, route) are not working in `options` 


dafortin:

sglanzer-fire:
Fixed in #337

PR #337, opened 2016-12-02
Fixing the spread for frost-link

sglanzer-fire:

Addresses #333 

# CHANGELOG

* Updated frost-link to accept named parameters in addition to position parameters, which unblocks the `spread` use case


addon/components/frost-link.js
```diff
@@ -243,11 +243,46 @@ export default LinkComponent.extend(SpreadMixin, PropTypeMixin, {
...
+      params.push(this.get('route'))
+
+      const queryParams = this.get('queryParams')
+      if (isPresent(queryParams)) {
+        params.push(queryParams)
+      }
+
+      const models = this.get('models')
+      if (!isEmpty(models)) {
+        params.push(this.get('models'))
```
quincyle: We should expect models to be an array instead of an object, so probably do a forEach/push here? Models should also come before queryParams (text, route, models, queryParams)https://github.com/emberjs/ember.js/blob/v2.8.0/packages/ember-htmlbars/lib/components/link-to.js#L805

addon/components/frost-link.js
```diff
@@ -243,11 +243,46 @@ export default LinkComponent.extend(SpreadMixin, PropTypeMixin, {
...
+   * https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-glimmer/lib/components/link-to.js#L790
+   *
+   * This allows named parameters to be passed to the link instead of positional
+   * params.  However, because the link-to expects positional params the named
+   * parameters must be set back against the component as a positional param
+   * array in the correct order (text, route, queryParams, models).  Once set
+   * we're free to hand control back to the parent function and it will react
+   * as if we used the original link-to interface.
+   */
+  didReceiveAttrs () {
```
quincyle, line 254: In ember 2.8 context, the htmlbar link-to component was still processing the positional params in **willRender** hook. It has been moved to **didReceiveAttrs** since 2.10 when ember switch from htmlbar to glimmar.
 **But I guess it will work fine since didReceiveAttrs runs before willRender anyway.**

quincyle (approved):

addon/components/frost-link.js
sglanzer-fire: It doesn't work that way - the code literally expects that to be an array 

https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-glimmer/lib/components/link-to.js#L834

https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-glimmer/lib/components/link-to.js#L758

Remember we aren't providing the positional params ourselves here, we're setting the equivalent result

sglanzer-fire:
@dafortin @rox163 getting timeouts in the tests?  I thought we had fixed the impact of ember-cli-code-coverage?

addon/components/frost-link.js
sglanzer-fire: Ah crud - you're totally right, I misread and forgot the order coming back to this a few days later - will fix

dafortin:
We 'skip' the tests that were causing these timeouts. The problem is probably the acceptance test you are adding. I think code coverage does not like our acceptance tests in general. That's weird I thought it was due to the specific set of tests we skipped.

sglanzer-fire:
Hmm, I can skip the acceptance test for now, but that's really something we need to fix, since it's impossible to check link behaviour without a running router.

dafortin:
I totally agree with you. We absolutely need to figure this one out.

dafortin (approved):

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/337/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #341, opened 2016-12-05
Add frost component

job13er:

# CHANGELOG

* **Added** `extends` helper to allow adding properties to a pojo (for use with `hookQualifiers`. Will add docs to `README.md` in next PR
* **Added** `ehook` helper to allow setting hookQualifiers as a POJO and not just key/value pairs (only useful until [this pr](https://github.com/Ticketfly/ember-hook/pull/35) is merged). Will add docs to `README.md` in next PR
* **Added** `CssMixin` - will add docs in `README.md` in next PR
* **Added** `Component` base class `addon/components/frost-component.js` to easily include all desired/required mixins for our components (`CssMixin`, `HookMixin`, `PropTypesMixin`, and `SpreadMixin` currently).
 * **Updated** existing components to derive from `frost-component`


job13er:
NOTE: I have a couple other somewhat related changes that should follow soon:
 1. Updating `frost-text` and `frost-password` (and I think another) to use `hookPrefix` instead of `receivedHook` and to `deprecate` `isHookEmbedded` property. I'll include the docs on the new stuff I added here in this one. 
 1. Refactoring remaining component tests to use more of a `describe/beforeEach/it` BDD style
 1. Adding name-scoped CSS classes alongside existing ones in both templates and SASS, then deprecating the non-scoped classes.
 1. Refactoring `frost-link` to be based off of `frost-component` and also support external links via an `href` property

I've got most of the work for 2 and 4 done in different branches on my fork, so those shouldn't take too long, and the others are pretty straightforward, but I may need to take a couple days and do some other work before I get those PRs out.

package.json
```diff
@@ -43,7 +43,7 @@
     "ember-cli-inject-live-reload": "^1.4.0",
     "ember-cli-mirage": "~0.2.3",
     "ember-cli-mocha": "^0.11.0",
-    "ember-cli-notifications": "^4.0.8",
+    "ember-cli-notifications": "~4.0.8",
```
job13er, line 47: it appears `4.1.1` or one similar to that is what broke this. Took the easy way out and pinned this instead of ripping it out as I've run out of time to work on this today.

sglanzer-fire: Yeah, that was my instinct as well - works for me, don't know why we would need to upgrade anyway

job13er:
Not sure why bitHound is hanging (broken link in the Details) but they were green in previous builds

addon/mixins/css.js
```diff
@@ -0,0 +1,67 @@
...
+
+  // == Computed Properties ===================================================
+
+  // == Functions =============================================================
+
+  /**
+   * @returns {String} the name of the component, parsed out of the toString() result
+   */
+  getComponentName () {
+    return this.toString().replace(/^.+:(.+)::.+$/, '$1')
```
sglanzer-fire, line 42: Do we want to mark this as using a private function?  I don't know how we want to track those kinds of things

addon/mixins/css.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+
+  /**
+   * This function is really only here because we can't stub this._super
+   * @see {@link https://github.com/emberjs/ember.js/issues/12457}
+   * It let's us test the functionality in the `init()` method w/o having to call
```
sglanzer-fire: lets

addon/mixins/css.js
```diff
@@ -0,0 +1,67 @@
...
+   * the real this._super() which was breaking when testing with tagName === '' for some reason
```
sglanzer-fire, line 49: I believe blank tags aren't allowed to have attributes in the newer Ember versions (may be your root cause there)

addon/mixins/css.js
job13er: It's not anymore, `toString()` is public :)

job13er: The only "private" bit is the format of the `toString()` value, we might wanna call that out I guess.

addon/mixins/css.js
job13er: that's why I need the code there, it was throwing errors when the `tagName` was `''` and `classNameBindings` was set.

job13er: But it's not the error I got in the test, it was a hard to understand error deep somewhere (can't find `lookup` on `undefined` or something like that)

tests/unit/mixins/css-test.js
```diff
@@ -0,0 +1,84 @@
...
+const {Component} = Ember
+import PropTypesMixin from 'ember-prop-types'
+import {afterEach, beforeEach, describe, it} from 'mocha'
+import sinon from 'sinon'
+
+import CssMixin from 'ember-frost-core/mixins/css'
+
+const CssComponent = Component.extend(PropTypesMixin, CssMixin, {
+  toString () {
+    return '<app-name@component:component-name::ember555>'
```
sglanzer-fire, line 15: Any way we can pull this off a real component so that we catch any changes in Ember with a break in our tests?

sglanzer-fire (approved):
Looking good - thanks for all the work on conventions, this is looking really clean and terse

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/341/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

tests/unit/mixins/css-test.js
job13er: I tried (briefly) and ended up with something in a different format than it ends up in when we're in an app. Maybe if I register it it'll work better. Lemme give that a try.

job13er: I'll go ahead and merge this and submit a follow-up with that enhancement (if I figure it out)

PR #350, opened 2016-12-13
fix for hasBlock - frost-link

srowhani:

# CHANGELOG
- Add check for text to handle case where link is not inline before pushing it into positional params

### DIFF
--- a/addon/components/frost-link.js
+++ b/addon/components/frost-link.js
@@ -254,9 +254,9 @@ export default LinkComponent.extend(SpreadMixin, PropTypeMixin, {
   didReceiveAttrs () {
     if (isEmpty(this.get('params'))) {
       let params = []
-
-      if (this.get('text')) {
-        params.push(this.get('text')) 
+      const text = this.get('text')
+      if (text) {
+        params.push(text) 
       }
 
       params.push(this.get('route'))
--- a/addon/components/frost-link.js
+++ b/addon/components/frost-link.js
@@ -255,7 +255,9 @@ export default LinkComponent.extend(SpreadMixin, PropTypeMixin, {
     if (isEmpty(this.get('params'))) {
       let params = []
 
-      params.push(this.get('text'))
+      if (this.get('text')) {
+        params.push(this.get('text')) 
+      }
 
       params.push(this.get('route'))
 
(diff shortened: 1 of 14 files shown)


addon/components/frost-link.js
```diff
@@ -255,7 +255,9 @@ export default LinkComponent.extend(SpreadMixin, PropTypeMixin, {
     if (isEmpty(this.get('params'))) {
       let params = []
 
-      params.push(this.get('text'))
+      if (this.get('text')) {
```
dafortin: Can should probably store this.get('text') in a variable to avoid getting it twice:
```
const text = this.get('text')
if (text) {
  ...
}
```

srowhani: For sure

sglanzer-fire:
So to clarify, is this expected to cover a case where there block usage for the text but named params otherwise?

```
{{#frost-link
  route='foo'
}}
  Some text
{{/frost-link}}
```

srowhani:
Yea if the component `hasBlock`, it will use that as it's `text` property. The link-to helper will consider the first positional param argument as the `route`.

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/350/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
Looks like there are some general failures in the build unfortunately

test-support/helpers/ember-frost-core.js
```diff
@@ -3,7 +3,15 @@ import Ember from 'ember'
 const {$, typeOf} = Ember
 import {$hook} from 'ember-hook'
 
-const assign = Object.assign || Ember.assign || Ember.merge
+const assign = Object.assign || Ember.assign || Ember.merge || function (assignee, assigner) {
```
job13er: I don't think you want/need to have a polyfill for assign here. It was failing because of lack of support for ember 2.3 in ember-test-utils, which I have since fixed.

job13er (changes requested):

test-support/helpers/ember-frost-core.js
job13er: https://github.com/ciena-blueplanet/ember-test-utils/pull/15/files

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/350/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

Issue #257, opened 2016-11-09
frost-checkbox: Setting autofocus property to true does not focus the element


juwara0:
PROBLEM DESCRIPTION:
Setting the autofocus=true does not cause the checkbox to be autofocuses. Here is the relevant section of code: https://github.com/juwara0/ember-frost-core/blob/f47a78880d64ab975d5b9c5d4717fcf54c4525e2/addon/components/frost-checkbox.js#L73-L79

This has something to do with setting the display: none on the input: https://github.com/juwara0/ember-frost-core/blob/f47a78880d64ab975d5b9c5d4717fcf54c4525e2/addon/styles/_frost-checkbox.scss#L55

When the display: none is removed autofocus works as expected. This is not a fix though since we want to use <svg> instead for styling purposes.

TASK DESCRIPTION: 
Find a way to make the autofocus=true work when using <svg>
Update the autofocus integration test to validate that autofocus is working correctly.

juwara0:
I investigated two possible solutions:
1. Attach the focus to the <label> tag - I ran into problems with this approach in that this worked find in Chrome but not in Firefox.

```
didInsertElement () {
    if (this.get('autofocus')) {
      console.log('autofocus called')
      run.next('render', () => {
        this.$('label').focus()
      })
    }
  },
```
1. Set the CSS from display: none to instead make the input very small: 1px wide by 1px high

```
.input[type=checkbox] {
      position: absolute;
      top: 2px;
      left: 2px;
      outline: none;
      width: 1px; height: 1px;
  }
```

This approach allowed the focus to work as expected in both Chrome and Firefox but tabbing resulted in 2 tabs instead of just one to move through the element.

srowhani:
@juwara0 Another approach could be passing a custom class to the `this.render` block so that `display != none`.

PR #358, opened 2016-12-30
Perform minor cleanup

sandersky:

# CHANGELOG

* Performed minor cleanup with `npm run eslint --- --fix` while running latest [eslint-config-frost-standard](https://github.com/ciena-frost/eslint-config-frost-standard) which includes new rules from [eslint-plugin-ember-standard](https://github.com/ciena-blueplanet/eslint-plugin-ember-standard) and [eslint-plugin-ocd](https://github.com/ciena-blueplanet/eslint-plugin-ocd).


sglanzer-fire (approved):
Love the fact that this was done via the fix command - stellar work!

sglanzer-fire:
Approved 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/358/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-select-dropdown.js
```diff
@@ -91,7 +91,7 @@ export default Component.extend({
     ]
       .join(';')
 
-    return Ember.String.htmlSafe(style)
+    return EmberString.htmlSafe(style)
```
juwara0: Since we do not have more than one String library in use right now, I would rather see us keep this as `String.htmlSafe()` and destructuring as `const {String} = Ember`.

addon/components/frost-select-dropdown.js
```diff
@@ -108,7 +108,7 @@ export default Component.extend({
       style.push(`bottom:${bottom - ARROW_HEIGHT + BORDER_HEIGHT}px`)
     }
 
-    return Ember.String.htmlSafe(style.join(';'))
+    return EmberString.htmlSafe(style.join(';'))
```
juwara0: Same comment as above.

addon/components/frost-select-li.js
```diff
@@ -48,7 +48,7 @@ export default Component.extend({
       const pattern = new RegExp(filter, 'gi')
       const label = data.label.replace(pattern, '<u>$&</u>')
 
-      return Ember.String.htmlSafe(label)
+      return EmberString.htmlSafe(label)
```
juwara0: Same as previous comment.

tests/dummy/app/pods/field/controller.js
```diff
@@ -1,10 +1,10 @@
 import Ember from 'ember'
 const {
-  Controller
+  Controller, inject
```
juwara0: We should update this to be on one line to follow our new conventions.

tests/dummy/app/pods/icons/controller.js
```diff
@@ -16,7 +16,7 @@ export default Controller.extend({
   iconPacks: computed('iconPacks', () => {
     return Object.keys(iconPacks).map((name) => {
       return {
-        name: Ember.String.capitalize(name),
+        name: EmberString.capitalize(name),
```
juwara0: Same comment as above about using String.

tests/dummy/app/pods/radio/controller.js
```diff
@@ -15,7 +15,7 @@ export default Controller.extend({
     {label: 'e', value: 'e'}
   ],
 
-  model: Ember.Object.create({
+  model: EmberObject.create({
```
juwara0: I would prefer that we not do this since we only have one Object library right now.

tests/dummy/app/router.js
```diff
@@ -1,7 +1,8 @@
-import Ember from 'ember'
 import config from './config/environment'
+import Ember from 'ember'
+const {Router: EmberRouter} = Ember
 
-var Router = Ember.Router.extend({
+var Router = EmberRouter.extend({
```
juwara0, line 5: I would prefer that we not do this since we only have one Router library right now.

tests/integration/components/frost-password-test.js
```diff
@@ -1,15 +1,15 @@
 import {expect} from 'chai'
 import {
+  $hook
```
juwara0: We can collapse this to one line to follow the new convention.

tests/integration/components/frost-toggle-test.js
```diff
@@ -1,15 +1,15 @@
 import {expect} from 'chai'
 import {
+  $hook
+} from 'ember-hook'
+import {
```
juwara0: We can collapse these to one line to follow the new convention.

tests/unit/components/frost-radio-button-test.js
```diff
@@ -1,17 +1,17 @@
...
+import Component from 'ember-frost-core/components/frost-component'
 import { describeComponent } from 'ember-mocha'
 import {
   beforeEach,
   describe,
   it
 } from 'mocha'
 
-import Component from 'ember-frost-core/components/frost-component'
+const {
```
juwara0: We can collapse this to one line to follow the new convention.

tests/unit/components/frost-radio-group-test.js
```diff
@@ -1,17 +1,17 @@
...
 import {
   beforeEach,
   describe,
   it
 } from 'mocha'
 
-import Component from 'ember-frost-core/components/frost-component'
+const {
+  run
+} = Ember
```
juwara0: We can collapse this to one line.

tests/unit/helpers/array-test.js
```diff
@@ -1,11 +1,11 @@
 import { expect } from 'chai'
 import {
+  array
+} from 'ember-frost-core/helpers/array'
+import {
```
juwara0: We can collapse these to one line.

tests/unit/utils/utils-test.js
```diff
@@ -1,13 +1,14 @@
...
+import {cloneEvent} from 'ember-frost-core/utils'
 import {
   describe,
   it
 } from 'mocha'
-import {cloneEvent} from 'ember-frost-core/utils'
+
+const {
+  $
+} = Ember
```
juwara0: We can collapse the import and const to one line.

addon/helpers/ehook.js
```diff
@@ -4,7 +4,7 @@
  * @see {@link https://github.com/Ticketfly/ember-hook/pull/35}
  */
 import Ember from 'ember'
-const {Helper, deprecate} = Ember
+const {Helper, deprecate, keys} = Ember
```
notmessenger, line 7: Do we have guidance about whether these should be alphabetized?

tests/acceptance/link-test.js
```diff
@@ -1,7 +1,7 @@
-import { describe, it, beforeEach, afterEach } from 'mocha'
-import { expect } from 'chai'
-import startApp from '../helpers/start-app'
 import destroyApp from '../helpers/destroy-app'
+import startApp from '../helpers/start-app'
+import { expect } from 'chai'
```
notmessenger: Are there supposed to be spaces before/after the braces?

tests/acceptance/link-test.js
```diff
@@ -1,7 +1,7 @@
...
+import { afterEach, beforeEach, describe, it } from 'mocha'
```
notmessenger: Are there supposed to be spaces before/after the braces?

addon/helpers/ehook.js
notmessenger: The result of this inquiry applies equally to many other changes in this PR in other files.

addon/components/frost-select-dropdown.js
sandersky: String is a Javascript class that is globally available. If we import as String we override that which could have unforeseen side effects.

tests/dummy/app/pods/field/controller.js
sandersky: I agree, but that is another ESLint rule we should create (this stuff was automatically done by running `npm run eslint --- --fix`). Anything else we want done I want to create more rules for so we implement the fix once and get it everywhere for free.

tests/dummy/app/pods/radio/controller.js
sandersky: Same as with `String`, `Object` is a Javascript global.

tests/dummy/app/router.js
sandersky: Notice the variable here is named `Router`. Either the import or this variable which is exported needs renamed.

addon/helpers/ehook.js
sandersky: They actually are alphabetized. When you sort an array of strings uppercase letters all come before lowercase. If we disagree with that then we need to update the sorting rules in `eslint-plugin-ocd`.

tests/integration/components/frost-password-test.js
notmessenger: Same with the `ember-mocha` line below as well?

tests/integration/components/frost-toggle-test.js
notmessenger: Same with the `mocha` line below?

tests/unit/components/frost-button-test.js
```diff
@@ -1,6 +1,6 @@
 import {expect} from 'chai'
 import Ember from 'ember'
-const {run} = Ember
+import Component from 'ember-frost-core/components/frost-component'
 import {describeComponent} from 'ember-mocha'
 import {
```
notmessenger: Should be on one line?

tests/unit/components/frost-checkbox-test.js
```diff
@@ -1,6 +1,6 @@
 import {expect} from 'chai'
 import Ember from 'ember'
-const {$, run} = Ember
+import Component from 'ember-frost-core/components/frost-component'
 import {describeComponent} from 'ember-mocha'
 import {
```
notmessenger: Should these all be one line?

tests/unit/components/frost-icon-test.js
```diff
@@ -1,16 +1,16 @@
 import {expect} from 'chai'
 import Ember from 'ember'
-const {run} = Ember
+import Component from 'ember-frost-core/components/frost-component'
 import {describeComponent} from 'ember-mocha'
+import PropTypeMixin from 'ember-prop-types'
+import SpreadMixin from 'ember-spread'
 import {
```
notmessenger: Should these all be one line?

tests/unit/components/frost-link-test.js
```diff
@@ -1,16 +1,17 @@
 import {expect} from 'chai'
 import Ember from 'ember'
-const {run} = Ember
 import {describeComponent} from 'ember-mocha'
+import PropTypeMixin from 'ember-prop-types'
+import SpreadMixin from 'ember-spread'
 import {
```
notmessenger: Should these all be one line?

tests/unit/components/frost-radio-group-test.js
```diff
@@ -1,17 +1,17 @@
 /* jshint expr:true */
 import { expect } from 'chai'
 import Ember from 'ember'
-const {
-  run
-} = Ember
+import Component from 'ember-frost-core/components/frost-component'
 import { describeComponent } from 'ember-mocha'
 import {
```
notmessenger: Should these all be one line?

addon/helpers/ehook.js
notmessenger: yeah, didn't think about that occurrence

test-support/helpers/ember-frost-core.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable ember-standard/destructure */
```
notmessenger, line 1: Why is this added to some file but not others?  I have not yet discovered the pattern requiring it.

addon/components/frost-select-dropdown.js
job13er: That's the same reason we don't do
```js
const {Object} = Ember
```
Even if there are no unforseen side-effects, I think it's misleading for things like `Object`, `String`, `Number` to be overwritten within a module. You probably wouldn't want a variable called `window` for a similar reason.

addon/components/frost-select-dropdown.js
job13er: Same justification as above.

addon/components/frost-select-li.js
job13er: Ditto.

tests/dummy/app/router.js
job13er: I would rename the instance personally.

tests/acceptance/link-test.js
job13er: I would vote no. Our current eslint rule only enforces consistency, i.e. `import {expect } from 'chai'` will fail, but I've been removing the extra padding spaces when I see them (they were generated that way but in our new generators I'm not adding the extra spaces).

addon/components/frost-select-dropdown.js
juwara0: Alright, then why not just leave the invocation as `Ember.String.htmlSafe()` instead of trying to destructure and having to assign it to an alias.

sandersky: @juwara0 one advantage is if you make a typo here:
```js
import Ember from 'ember'
const {Component, Strung: EmberString} = Ember
import computed, {readOnly} from 'ember-computed-decorators'

export default Component.extend({
  @readOnly
  @computed('title')
  header (title) {
    return EmberString.htmlSafe(`<h1>${title}</h1>`)
  }
})
```
it will be caught at compile time whereas the following typo won't:
```js
import Ember from 'ember'
const {Component} = Ember
import computed, {readOnly} from 'ember-computed-decorators'

export default Component.extend({
  @readOnly
  @computed('title')
  header (title) {
    return Ember.Strung.htmlSafe(`<h1>${title}</h1>`)
  }
})
```

tests/acceptance/link-test.js
sandersky: I would vote no as well but this is yet to be an ESLint in our stack so until it is shouldn't block PR's.

tests/dummy/app/router.js
juwara0: Or we just leave this as `Ember.Router`.

tests/acceptance/link-test.js
sandersky: If we all agree then we can simply add http://eslint.org/docs/rules/object-curly-spacing to eslint-config-frost-standard.

addon/components/frost-select-dropdown.js
juwara0: I am not crazy about the `EmberString`, `EmberObject` and `EmberRouter` approach. It feels forced just so that we can destructure but then the need to use an alias makes it less clean. I think that in these cases it is more clear to just use `Ember.String`, `Ember.Object` and `Ember.Router`.

job13er: I agree with the first two, the second should never be needed, there shouldn't be another `Router` symbol in that file, it should be renamed. 

I believe the agreed upon approach here is to begin a discussion in `#frost-desgin` if we want to move away from the previous agreed-upon convention of descructuring everything *except* what clashes with a built-in and leaving those on the `Ember` object `Ember.String`, `Ember.Object` but *not* `Ember.Router`, in that case, the `var Router` should be changed (in case I didn't say that enough times yet ;)

test-support/helpers/ember-frost-core.js
sandersky: The new destructure rule wants all Ember classes destructured except String and Object. In the case of this file it does:
```js
const assign = Object.assign || Ember.assign || Ember.merge
```
in order to support older versions of Ember. Without this rule disabled it would complain that `Ember.assign` and `Ember.merge` are not destructured.

tests/integration/components/frost-checkbox-test.js
```diff
@@ -248,7 +245,7 @@ describeComponent(
     })
 
     // https://github.com/juwara0/ember-frost-core/issues/1
-    it.skip('sets focus on render when autofocus is true', function () {
+    it('sets focus on render when autofocus is true', function () {
```
juwara0: The issue that is causing this test to fail is not fixed yet: https://github.com/ciena-frost/ember-frost-core/issues/257 

This test was skipped until the issue was fixed.

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/358/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #376, opened 2017-01-13
Added min/max properties to frost-text

TheOtherDude:

# CHANGELOG

* Added min/max properties to `frost-text` component

### DIFF
--- a/addon/components/frost-text.js
+++ b/addon/components/frost-text.js
@@ -29,6 +29,8 @@ export default Component.extend(FrostEventsProxyMixin, {
     isClearEnabled: PropTypes.bool,
     isClearVisible: PropTypes.bool,
     isHookEmbedded: PropTypes.bool,
+    max: PropTypes.number,
+    min: PropTypes.number,
     receivedHook: PropTypes.string,
     tabindex: PropTypes.number,
     type: PropTypes.string,
--- a/addon/templates/components/frost-text.hbs
+++ b/addon/templates/components/frost-text.hbs
@@ -8,7 +8,9 @@
   disabled=disabled
   form=form
   hook=(concat receivedHook '-input')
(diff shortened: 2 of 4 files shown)


sandersky:
@TheOtherDude can you add an example on the dummy app so other devs can easily discover the feature?

TheOtherDude:
Have we decided to go ahead with this implementation? I was assuming it was DOA. If not, I can definitely add an example.

sandersky:
I don't know if it is DOA or not. If it isn't then I'd like to see an example, otherwise the PR should be closed.

sglanzer-fire:
I assume this is contingent on the discussion around blocking invalid input vs erroring invalid input?  Sorry, I've been completely dead to the world on app upgrades for the last week

TheOtherDude:
@sglanzer Correct. Based on our slack discussion, it sounded like we wanted to handle input outside the given bounds with an `onError` action rather than passing the bounds down to the raw input and letting the browser take care of it.

However, UX later came in and said allowing the browser to block invalid input might be acceptable, but they would need to discuss it amongst themselves. So far, I have not heard what came out of those discussions.

TheOtherDude:
Just wondering if it there has been any word from UX on this. As it stands, we're using a raw input styled like a `frost-text` because we need this functionality (according to our resident UX expert).

sglanzer-fire:
Having seen the numeric version of the input again this week, I'm heavily inclined towards making a numeric input a separate component with optional controls.  The current version doesn't look very good and I'd like to be able to handle cases like this with min/max in a more visible way to consumers

I've added a feature request to the project to track https://github.com/ciena-frost/ember-frost-core/issues/405

TheOtherDude:
Sounds good!

PR #377, opened 2017-01-18
Added support for multiple routes

dafortin:

With the current implementation of `frost-link` we can open multiple tabs with different route names but not with different models or query params. I'm proposing this new implementation to be able to have different routes with different models and query params.

Current implementation:
```
{{frost-link 'Text'
      hook='myMultipleRoutesLink'
      priority='primary'
      routeNames=(array
        'link.min'
        'link.max'
      )
      size='small'
}}
```

New feature:
```
{{frost-link 'Text'
      hook='myMultipleRoutesLink'
      priority='primary'
      routes=(array
        (hash
          name= 'link.first'
          models= (array '1')
        )
        (hash
          name= 'link.first.second'
          models= (array '2')
        )
      )
      size='small'
}}
```

Note: 
* This is a proposition. Feel free to comment or reject.
* <b>There is no test in this PR<b>. I can raise an issue to add those. I will take care of it once our current task is done.

# CHANGELOG
* Added support for multiple routes for `frost-link`

### DIFF
--- a/addon/components/frost-link.js
+++ b/addon/components/frost-link.js
@@ -7,7 +7,7 @@ import computed, {readOnly} from 'ember-computed-decorators'
 import {HookMixin} from 'ember-hook'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
 import SpreadMixin from 'ember-spread'
-const {LinkComponent, Logger, deprecate, get, isEmpty, isPresent, run, set} = Ember
+const {LinkComponent, Logger, deprecate, isEmpty, isPresent, run, set} = Ember
 
 /**
  * List of valid values to pass into `design` propery
@@ -172,7 +172,7 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
    * @returns {boolean} true if we should open the link in the current tab and false otherwise
    */
   _shouldOpenInSameTab () {
-    return !(get(this, 'priority') === 'primary' && get(this, 'disabled') === false)
+    return !(this.get('priority') === 'primary' && this.get('disabled') === false)
   },
 
   /**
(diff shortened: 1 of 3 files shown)


addon/components/frost-link.js
```diff
@@ -71,6 +71,11 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
     icon: PropTypes.string,
     priority: PropTypes.oneOf(validPriorities),
     routeNames: PropTypes.array,
```
sglanzer-fire, line 73: I assume routeNames would be deprecated now?

addon/components/frost-link.js
```diff
@@ -174,30 +180,49 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
    * @returns {boolean} true if we need to open multiple links on click and false otherwise
    */
   _hasMultipleLinks () {
-    return get(this, 'routeNames') !== undefined && get(this, 'routeNames').length !== 0
+    return (get(this, 'routeNames') !== undefined && get(this, 'routeNames').length !== 0) ||
```
sglanzer-fire: The current convention for `get` against a component property is `this.get` - we're reserving `get(object, 'prop')` for non-`this` cases

addon/components/frost-link.js
```diff
@@ -174,30 +180,49 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
...
-    if (routeNames) {
       routeNames.forEach((routeName) => {
-        const windowHandler = window.open(routing.generateURL(routeName, models, queryParams))
-        if (!windowHandler) {
-          Logger.warn('Warning: Make sure that the pop-ups are not blocked')
-        }
+        this._openLink(routeName, models, queryParams)
+      })
+    } else if (!isEmpty(routes)) {
+      routes.forEach((route) => {
```
sglanzer-fire: Just for kicks - you can destructure `route` in this case as
```
routes.forEach(({name, models, queryParams}) => {
  this._openLink(name, models, queryParams)
})
```

addon/components/frost-link.js
```diff
@@ -213,14 +238,26 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
...
       } else {
         set(this, 'target', '_blank')
       }
     }
   },
+
+  _warnPropertyPrecedence () {
+    let attributeName
+    if (get(this, 'routeName')) {
+      attributeName = 'routeName'
```
sglanzer-fire: I don't think you meant this to be `routeName`

sglanzer-fire:
Good overall, I'd just like some quick cleanup and we can merge

addon/components/frost-link.js
dafortin: What is the convention for deprecation? Should we only add a comment or is there anything else to do?

job13er: You need to check for it being assigned, probably in `didReceiveAttrs` and if it is, call `deprecate` (from `Ember`). If the `routeNames` paran still functions as expected, then that should be all needed. If not, you'll need to take what was given in `routeNames` and convert it to whatever is necessary to achieve the same as what `routeNames` used to do

dafortin:
Add a task to add test for this new feature: #378

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/377/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
Approved

PR #379, opened 2017-01-19
Fixed code to work with Ember 2.10

sandersky:

Tested in a real app without issue. The test results from that app are:
```
ran 2003 tests in 6.86835 min [1996 passed / 0 failed / 7 pending]
```

For those concerned with my removal of *tests/unit/components/frost-link-test.js* I can assure you I've made sure to add the right integration tests as a replacement. The unit tests in this file simply won't work with Ember 2.10 as the **Ember.LinkComponent** NEEDS positional parameters to not throw an error in tests, while the *frost-link* extension doesn't require them. Here is a screenshot of the coverage for the *frost-link* component with this PR's changes, showing it is well tested.

screen shot 2017-01-19 at 2 53 07 pm

Below are the things not covered, which are edge cases that I don't think we can actually even enter to be honest. With that being said I'd even be open to removing these extra checks if desired.

screen shot 2017-01-19 at 2 56 54 pm

> Note: In the above case a valid *frost-link* must contain the route as a positional param, via the **routeNames** option, or via the **routes** option. This *else* that isn't entered would be if none of those three were provided, at which point you wouldn't be using the *frost-link* API correctly to begin with.

---

screen shot 2017-01-19 at 2 57 05 pm

> Note: The above method is only called when **routeNames** or **routes** is present and thus the *else* condition is never entered.

---

screen shot 2017-01-19 at 2 57 08 pm

> Note: The above if condition can probably go away as we never call *_openLink()* without a route name.

---

I also fixed a bug while I was performing this cleanup which you can see in the screenshot from the currently published demo app (notice the two links on the right are missing the text):

screen shot 2017-01-19 at 3 12 07 pm

# CHANGELOG

* **Fixed** bug in `frost-link` where text wasn't showing up when using `routes` or `routeNames` options while using an inline `frost-link`.
* **Fixed** code so that this addon will work in Ember 2.10 as well as versions dating all the way back to Ember 2.3.


sglanzer-fire:
Approved 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/379/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #380, opened 2017-01-20
Fix select in demo

sandersky:

I've also integration tested this fix against an app running on Ember 2.8 to make sure the addition of the `run.ext` call within the `frost-select` didn't cause any issues.

# CHANGELOG

* **Addedd** missing acceptance tests to visit remaining routes in demo app to ensure all pages continue to load without issue.
* **Fixed** deprecation warning about modifying property twice in one render (which became an error in Ember 2.10).
* **Fixed** select route in demo.

### DIFF
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -368,7 +368,9 @@ export default Component.extend({
       const onChange = this.get('onChange')
 
       if (typeOf(onChange) === 'function') {
-        this.onChange(selectedValue)
+        run.next(() => {
+          this.onChange(selectedValue)
+        })
       }
 
       // We need to make sure focus goes back to select since it is on the
--- a/tests/acceptance/application-test.js
+++ b/tests/acceptance/application-test.js
@@ -8,6 +8,7 @@ describe('Acceptance: Application', function () {
 
   before(function () {
     application = startApp()
+    server.loadFixtures()
   })
 
   after(function () {
@@ -29,10 +30,16 @@ describe('Acceptance: Application', function () {
     'button',
     'checkbox',
     'field',
+    'helpers',
     'icons',
     'layout',
+    'loading',
     'palette',
     'password',
+    'radio',
+    'scroll',
+    'select',
+    'toggle',
     'typography'
   ]
     .forEach((path) => {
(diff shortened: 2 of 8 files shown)


addon/components/frost-select.js
```diff
@@ -345,7 +345,9 @@ export default Component.extend({
       const onInput = this.get('onInput')
 
       if (typeOf(onInput) === 'function') {
-        onInput(filter)
+        run.later(() => {
```
sglanzer-fire: `later` or `next`?

sglanzer-fire: http://emberjs.com/api/classes/Ember.run.html#method_next

sandersky: They are essentially the same thing but I'll change it.

sandersky: It looks like using `run.next` instead actually causes some tests to fail where `run.later` doesn't.

addon/components/frost-select.js
```diff
@@ -368,7 +368,9 @@ export default Component.extend({
       const onChange = this.get('onChange')
 
       if (typeOf(onChange) === 'function') {
-        this.onChange(selectedValue)
+        run.next(() => {
```
sandersky, line 371: This has been causing deprecation warnings for a while in Ember 2.8, Ember 2.10 actually throws a hard error so I was finally forced to address this.

tests/acceptance/application-test.js
```diff
@@ -8,6 +8,7 @@ describe('Acceptance: Application', function () {
 
   before(function () {
     application = startApp()
+    server.loadFixtures()
```
sandersky, line 11: In order for the select route to work we need Mirage mocks loaded.

tests/dummy/config/environment.js
```diff
@@ -41,6 +41,10 @@ module.exports = function (environment) {
     ENV.rootURL = '/'
     ENV.locationType = 'none'
 
+    ENV['ember-cli-mirage'] = {
```
sandersky, line 44: We need Mirage mocks in our test environment for acceptance tests.

sandersky:
:+1: @sglanzer self-merging to fix issue in select

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/380/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #386, opened 2017-01-25
Refactor test helpers and add new selectAtIndex frost-select helper

sandersky:

select-trimmed-text

# CHANGELOG

* **Added** ellipsis in middle of select item labels when the text is trimmed instead of at the end as well as show the full label on mouseover. Prior to this change there was no way to see the full label when it is too wide.

* **Added** new `open()` and `selectItemAtIndex()` test helpers for `frost-select` which can be used like so (in example the hook for the select being tested is `mySelect`):

  ```js
  import {
    expectWithState as expectSelectWithState,
    open,
    selectItemAtIndex
  } from 'dummy/tests/helpers/ember-frost-core/frost-select'

  // before below test can run you'd need to render whatever DOM is necessary for your test

  describe('when select opened', function () {
    beforeEach(function () {
      open('mySelect')
    })

    describe('when first item selected', function () {
      beforeEach(function (done) {
        selectItemAtIndex('mySelect', 0, done)
      })

      it('renders as expected', function () {
        expectSelectWithState('mySelect', {
          focused: true,
          opened: false,
          text: 'My first item'
        })
      })
    })
  })
  ```

### DIFF
--- a/addon/components/frost-select-dropdown.js
+++ b/addon/components/frost-select-dropdown.js
@@ -46,11 +46,17 @@ export default Component.extend({
     selectedItems: PropTypes.arrayOf(PropTypes.object),
 
     // state
-    bottom: PropTypes.number,
+    bottom: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.string
+    ]),
     focusedIndex: PropTypes.number,
     left: PropTypes.number,
     maxHeight: PropTypes.number,
-    top: PropTypes.number,
+    top: PropTypes.oneOfType([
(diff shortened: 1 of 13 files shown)


sglanzer-fire:
Approved - I didn't check _all_ the test helpers, but enough to get the gist - love what you did with the trim function

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/386/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

Issue #369, opened 2017-01-10
Don't use the full path for the `css` property binding (just use file name)


sglanzer-fire:
https://github.com/ciena-frost/ember-frost-core/blob/master/addon/mixins/css.js#L28

PR #393, opened 2017-02-07
getComponentName returns clean className in nested component

srowhani:

I wanted to iterate on how `getComponentName` would set classNames because nested components were getting pretty messy.

classname which would currently be `baz/foo-bar` becomes `foo-bar`

# CHANGELOG
- css mixin `getComponentName` returns clean className for nested components

### DIFF
--- a/tests/unit/mixins/css-nested-test.js
+++ b/tests/unit/mixins/css-nested-test.js
@@ -1,0 +1,41 @@
+/**
+ * Unit test for the CssMixin
+ */
+import {expect} from 'chai'
+import {setupComponentTest} from 'ember-mocha'
+import {afterEach, beforeEach, describe, it} from 'mocha'
+import sinon from 'sinon'
+
+describe('Unit / Mixin / css-nested', function () {
+  let component, sandbox
+  const expectedName = 'foo-bar'
+  setupComponentTest('baz/foo-bar')
+
+  beforeEach(function () {
+    sandbox = sinon.sandbox.create()
+    component = this.subject()
+  })
+
+  afterEach(function () {
+    sandbox.restore()
+  })
+
+  it('has classNameBindings set to include "css"', function () {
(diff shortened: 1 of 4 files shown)


tests/unit/mixins/css-test.js
```diff
@@ -37,6 +37,13 @@ describe('Unit / Mixin / css', function () {
     it('should strip the junk from toString()', function () {
       expect(componentName).to.equal('css-component')
     })
+    it('should extract a className from a nested component', function () {
+      component.toString = function () {
+        return '<baz@component:deep/nested/component/foo-bar::id>'
+      }
```
job13er: Can we maybe actually create a nested component and check that one? So then if `toString()` changes for those nested components we'll know about it here instead of in an app somewhere.

job13er (commented):
Looks great, just the one request to use a real nested component.

tests/unit/mixins/css-nested-test.js
```diff
@@ -0,0 +1,41 @@
...
+  })
+
+  describe('.getComponentName()', function () {
+    let componentName
+    beforeEach(function () {
+      componentName = component.getComponentName()
+    })
+
+    it('should extract a className from a nested component', function () {
+      expect(componentName).to.equal(expectedName)
```
job13er, line 38: Not sure why the expected name is specified up above, instead of right here, but that's not a big deal

job13er (approved):

job13er:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/393/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

job13er:
Fixes https://github.com/ciena-frost/ember-frost-core/issues/369

PR #394, opened 2017-02-08
#388 Tabbing with select

### LINKED ISSUE #388: Tabbing with select
It seems you can begin typing into the select dropdown filter and then tab out of the select, moving focus to other inputs/links on the page. However at this point the dropdown is still visible making it hard to close without using a mouse. We need to make sure hitting tab while in the dropdown closes the dropdown under all scenarios.

srowhani:

Fixes #388 

# CHANGELOG
- Pass action closure `closeDropDown` into `frost-select-dropdown`


job13er:
@srowhani Looks like some `frost-multi-select` tests are failing now

job13er:
@srowhani A bugfix it's a bugfix without a test. Please include a test that fails b/c of the bug and passes now that there's a fix.

job13er (changes requested):
@srowhani A bugfix it's a bugfix without a test. Please include a test that fails b/c of the bug and passes now that there's a fix.

sglanzer-fire:
Seems reasonable, but I agree with Adam that I'd like to see a test added

PR #399, opened 2017-02-11
Improve frost-select performance

### LINKED ISSUE #398: Frost select performance
I have a frost-select in my app with 200 options.

Currently, the click to open it takes about 330 ms with `1.7.6` and over 900 ms with `1.9.0`. On other machines, we have timed it at over 3.25 seconds with `1.9.0`.

Something added between `1.7.6` and `1.9.0` is really killing frost-select performance.

TheOtherDude:

Resolves #398 

Setting the text and title on the `frost-select-li` approximately tripled its rendering time (at least in Chrome with Ember 2.8.3). With this fix, my frost-selects are performing nearly as well as they did pre `1.8.0`. 

# CHANGELOG

* **Improved** frost-select performance

### DIFF
--- a/addon/utils/text.js
+++ b/addon/utils/text.js
@@ -1,3 +1,4 @@
+/* global jQuery */
 import Ember from 'ember'
 const {$} = Ember
 
@@ -71,8 +72,9 @@ export function trimDataToFit (text, width, font) {
  * Note: expects data-text attribute to be on element with full text
  *
  * @param {HTMLElement} element - HTML element to trim data within
+ * @param {jQuery} [$] - jQuery selector, can be stubbed for testing
  */
-export function trimLongDataInElement (element) {
+export function trimLongDataInElement (element, $ = jQuery) {
   let text = element.textContent
   let $element = $(element)
   let tooltip = ''
@@ -89,9 +91,8 @@ export function trimLongDataInElement (element) {
   if (textWidth > width) {
     text = trimDataToFit(text, width)
     tooltip = $(element).data('text')
+    $element
+      .text(text)
+      .prop('title', tooltip)
   }
-
-  $element
-    .text(text)
-    .prop('title', tooltip)
 }
--- a/tests/integration/components/frost-select-test.js
+++ b/tests/integration/components/frost-select-test.js
@@ -1213,6 +1213,48 @@ describe(test.label, function () {
       })
     })
 
+    describe('When label is too long to fit', function () {
+      beforeEach(function () {
+        this.set('data', [
+          {label: 'Baaaaaaaaaa baaaaaaaaaaaaaaa black sheep', value: 'sheep'}
+        ])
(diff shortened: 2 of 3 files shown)


addon/utils/text.js
```diff
@@ -89,9 +91,8 @@ export function trimLongDataInElement (element) {
   if (textWidth > width) {
     text = trimDataToFit(text, width)
     tooltip = $(element).data('text')
+    $element
```
sandersky, line 94: I think rather than putting this in the if condition you should check the new text value against what is currently in the DOM before updating the DOM. Otherwise if a select is renderer to be the width of its container and the browser window is shrunk then expanded the full text may need to become visible again.

addon/utils/text.js
```diff
@@ -71,8 +72,9 @@ export function trimDataToFit (text, width, font) {
  * Note: expects data-text attribute to be on element with full text
  *
  * @param {HTMLElement} element - HTML element to trim data within
+ * @param {jQuery} [$] - jQuery selector, can be stubbed for testing
```
sandersky, line 75: I'm not sure I get what you're doing here as the tests appear to stub Ember.$ not the global jQuery.

TheOtherDude: Unfortunately, for reasons unknown, setting the default for `$` as `$ = $` doesn't work - the dollar sign ends up undefined. 

However, `Ember.$ === jQuery`, so I was just using the global to avoid destructuring linting errors. 

I'm still trying to understand why `$ = $` isn't working. Using a function to return the destructured `$` works fine: `$ = get$()`.

TheOtherDude: I guess it's just that the parameter overwrites the `$` from outside its scope before it can be set to the default value. Renaming the `$` also works: `$2 = $`.

addon/utils/text.js
TheOtherDude: The `frost-select-li` re-renders each time the select is opened (and apparently even each time the hover effect changes in the list), so this isn't really an issue. I'll put in a test case to make sure resizing the select between opening and closing the list continues to work, though.

TheOtherDude: The new test does fail in Ember 2.3, but if you run the demo app under Ember 2.3, you will see that the `frost-select` does not really work at all, even if you completely disable the trim feature. 

As such, I am not sure how to proceed. Do we need to continue Ember 2.3 support? The test passes in Ember 2.4.

sglanzer-fire:
Sorry, I must be missing something - where is the actual fix in this?

TheOtherDude:
The fix is to only set the text and title on the element when the text has been trimmed, not every time.

TheOtherDude:
Closing this in favor of the more comprehensive [PR#400](https://github.com/ciena-frost/ember-frost-core/pull/400)

PR #407, opened 2017-02-23
Added .pr-bumper.json from other projects

TheOtherDude:

# CHANGELOG

* **Fixed** pr-bumper configuration
* From a previously unpublished PR:
  * **Added** wrapLabels attribute to frost-select, which disables the text trimming for frost-select option labels and allows them to wrap.


.pr-bumper.json
```diff
@@ -0,0 +1,46 @@
...
+        "pattern": "\\s+\"(ember\\-frost\\-\\S+)\"",
+        "url": "https://github.com/ciena-frost/${REPO_NAME}.git"
+      },
+      {
+        "pattern": "\\s+\"(frost\\-\\S+)\"",
+        "url": "https://bitbucket.ciena.com/scm/bp_frost/${REPO_NAME}.git"
+      }
+    ]
+  },
+  "dependencySnapshotFile": "",
```
job13er: This should be the only line necessary to disable the dependency snapshot and get it working. The rest is explicitly setting defaults, which seems unnecessary.

job13er (changes requested):

.pr-bumper.json
TheOtherDude: Does the dependency snapshot require typescript? I'd like to understand why this issue is occurring because it's popping up in every project with `pr-bumper`.

PR #412, opened 2017-03-01
added debounceInterval into frost-select

### LINKED ISSUE #411: Add debounce to `select` when user is typing
Would be nice to have a debounce when the user is typing, especially when a server api is invoked, so that it's not going off to query on every character entered.

srowhani:

Closes #411 

# CHANGELOG
- Added `debounceInterval` into `frost-select`


addon/components/frost-select.js
```diff
@@ -119,7 +120,7 @@ export default Component.extend({
       renderTarget: 'frost-select',
       role: 'button',
       tabIndex: 0,
-
+      debounceInterval: 0,
```
sglanzer-fire, line 125: I'd default to at least 50

addon/components/frost-select.js
```diff
@@ -364,7 +365,8 @@ export default Component.extend({
       const onInput = this.get('onInput')
 
       if (typeOf(onInput) === 'function') {
-        onInput(filter)
+        const debounceInterval = this.get('debounceInterval')
+        run.debounce(this, onInput, filter, debounceInterval)
```
sglanzer-fire: Do we want to use ember-concurrency here or will the ember run automatically clean up as well?

tests/unit/components/frost-select-test.js
```diff
@@ -0,0 +1,48 @@
...
+      component.actions.filterInput.call(component, {
+        target: {
+          value: 'hello'
+        }
+      })
+      expect(onInput.called).to.equal(false)
+      run.later(function () {
+        expect(onInput.called).to.equal(true)
+        done()
+      }, 250)
```
sglanzer-fire: Maybe add another check at 150?

sglanzer-fire (changes requested):
Default and ember-concurrency question

sandersky:
This should be a #minor# not a #patch# as it is adding a feature rather than fixing a bug.

addon/components/frost-select.js
sandersky: @sglanzer if we don't make the default 0 don't we risk breaking consumers who could theoretically rely on every keystroke? I know some of my use cases rely on client-side filtering and it is nice having the instant feedback loop as the user types.

tests/unit/components/frost-select-test.js
```diff
@@ -0,0 +1,48 @@
+/**
```
sandersky: I'd personally prefer to see an integration test instead.

addon/components/frost-select.js
srowhani: `debounce` will essentially start a setTimeout. If debounced method is fired again, most recent will be set as the one to fire, and the previous call will be cancelled and popped from a queue.

Here's the internal implementation, if you are interested.
https://github.com/ebryn/backburner.js/blob/59cfe851ca0ce9b2e6c7be61435e22fab3731bdb/lib/backburner/index.js#L468

addon/components/frost-select.js
sglanzer-fire: It would effectively be instant as soon as they stop typing - fine, for backward compatibility we can keep it at 0

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/412/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
@d-ashfield this is the PR I was mentioning

sglanzer-fire (approved):

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/412/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-select.js
sandersky: I'm happy to change the default on the next major rev if we want.

sandersky:
:+1:

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/412/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #417, opened 2017-03-20
Added support for external link

dafortin:

# CHANGELOG

* **Added** support for external link

### DIFF
--- a/addon/components/frost-link.js
+++ b/addon/components/frost-link.js
@@ -154,9 +154,6 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
     hook: PropTypes.string.isRequired,
     hookPrefix: PropTypes.string,
     hookQualifiers: PropTypes.object,
-    // The user can define href for external links, but we don't want to define in propTypes because the CP will get
-    // overwritten even in the non external link use case.
-    // href: PropTypes.string,
     icon: PropTypes.string,
     linkTitle: PropTypes.string,
     onClick: PropTypes.func,
@@ -277,16 +274,6 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
   },
 
   /**
-   * Returns true if is an external link and false otherwise.
-   * @returns {boolean} true if is an external link and false otherwise
-   */
-  _isExternal () {
-    // We are expecting to get a CP is href was not specified by the component consumer.
-    const href = this.href
-    return !isEmpty(href) && typeOf(href) === 'string'
-  },
-
-  /**
    * Open multiple links.
    */
   _openLinks () {
@@ -321,18 +308,6 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
       if (!windowHandler) {
         Logger.warn('Warning: Make sure that the pop-ups are not blocked')
       }
-    }
-  },
-
-  /**
-   * Open a link for an external location.
-   */
-  _openExternalLink () {
-    const href = this.get('href')
-    // If we want to open a new tab, the target will be set accordingly when we setup the routing.
-    // If we want to stay in the same tab, we will set the href.
-    if (this._shouldOpenInSameTab()) {
-      window.location.href = href
     }
   },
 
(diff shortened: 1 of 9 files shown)


addon/components/frost-link.js
```diff
@@ -258,7 +264,7 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
    * @returns {boolean} true if we should open the link in the current tab and false otherwise
    */
   _shouldOpenInSameTab () {
-    return !(this.get('priority') === 'primary' && this.get('disabled') === false)
+    return this.get('priority') !== 'primary'
```
dafortin, line 264: There was a bug here. We were able to click even if the link was disabled. I fixed it and added a test.

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/417/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-link.js
```diff
@@ -309,6 +325,18 @@ export default LinkComponent.extend(PropTypeMixin, HookMixin, SpreadMixin, {
...
 
   /**
+   * Open a link for an external location.
+   */
+  _openExternalLink () {
+    const href = this.get('href')
+    // If we want to open a new tab, the target will be set accordingly when we setup the routing.
+    // If we want to stay in the same tab, we will set the href.
+    if (this._shouldOpenInSameTab()) {
+      window.location.href = href
```
job13er: Doesn't implementing it this way prevent the user from choosing, at the time of clicking the link, whether to open it in the same tab or in another tab? I'm not a big fan of links that redirect with JavaScript vs actually setting an `href` on an anchor tag. The latter is much more user friendly IMO. Instead of putting the decision of whether to open in the same or a new tab solely in the hands of the developer, it lets the user have some control over it. For example, I believe the UX on the links within an object browser are supposed to open in the same tab, but allow the user to open them in a new tab if desired. I don't think this implementation will allow that use-case.

job13er (dismissed):
I don't think this implementation supports the required use cases.

addon/components/frost-link.js
dafortin: Currently the implementation of the link is the following:
primary link: new tab
secondary link: same tab

I followed the same concept here.  

I don't think there is any way to open the link in the same tab if it's a primary link with the existing code.

dafortin: We can change the behavior but it will change the original behavior of the `frost-link`

job13er: I wasn't suggesting adding the ability to open a primary link in the same tab, but a secondary or tertiary link in a new tab. Just pointing out that without that functionality, `frost-link` is still unsusable for people who want/need to provide that functionality. 

Since I don't use `frost-link` now, I don't really care if you add this, just pointing out it doesn't solve the problem I raised about supporting external links in a way I can use.

job13er: Looks like I never even opened a ticket with that defect, so I'll shut up now and let you do whatever you want with a component I have no use for anyway, since it doesn't actually affect me at all.

dafortin: Ok thanks for explaining I get your point. It's actually a very valuable point. I will look into it.

sglanzer-fire: Yep - legit point - we should attempt to address this

dafortin:
Let's cancel this and do a `frost-external-link`. :)

sglanzer-fire:
As discussed, pretty much a clone of frost-link without the link-to specific portions.  Let's go with this for now as a shortest path - can you also raise an issue for the next major, I think we should look into creating our own link component (not extending Ember's) and we can roll these back together at that stage

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/417/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

dafortin:
As discussed - we will postpone this change to the next major release. We will prefer to do a single component handling both as described in #418

PR #423, opened 2017-03-23
Added blueprint helper and modified blueprint to check if packages already installed

dafortin:

# CHANGELOG
* **Added** blueprint helper
* **Modified** blueprint to check if package to install already exists


sglanzer-fire (changes requested):
Would you mind adding some comments (inline) about the strategy being used here and where we're overriding the default Ember-CLI behaviour?

Also, I don't think the title of the PR is accurate anymore

sandersky:
For starters I see no tests around the new code. Look at how I did it [here](https://github.com/ciena-blueplanet/ember-test-utils/blob/master/package.json#L24).

blueprints/ember-frost-core/index.js
```diff
@@ -1,35 +1,27 @@
...
-            {name: 'ember-frost-test', target: '^1.0.0'},
-            {name: 'ember-prop-types', target: '^3.0.0'},
-            {name: 'ember-truth-helpers', target: '^1.2.0'}
-          ]
-        })
-      })
+    // Get the packages installed in the consumer app/addon
+    const consumerPackages = blueprintHelper.consumer.getPackages(options)
+
+    // Get the packages to install (not already installed) from a list of potential packages
```
sglanzer-fire: Can we update this comment to add something like "Packages that are already installed in the consumer within the required semver range will not be re-installed or have blueprints re-run"?

Also might be worth noting this behaviour in the README - just a short blurb and point to the source (here)

lib/utils/consumer.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  /**
+   * Get the existing consumer dependencies.
+   * @param {Object} options the blueprint options
+   * @returns {Object} all the consumer dependencies
+   */
+  getPackages: function (options) {
+    return Object.assign(options.project.pkg.devDependencies, options.project.pkg.dependencies)
```
sglanzer-fire: I'd add a null guard here just in case the path is empty

lib/utils/npm.js
```diff
@@ -0,0 +1,53 @@
...
+const _Promise = require('promise')
+
+module.exports = {
+  /**
+   * Run an npm command.
+   * @param {String} command the name of the npm command (view, install, unistall, etc.)
+   * @param {Array} args a list of arguments
+   * @returns {Promise} a promise that will return the command result once it's run
+   */
+  _run: function (command, args = []) {
```
sglanzer-fire, line 14: Are any of these available from `ember-cli` modules to borrow from?

lib/utils/package-handler.js
```diff
@@ -0,0 +1,87 @@
...
+   * Get the packages to install. If a package is already installed, it will not required to be installed.
+   * @param {Array} pkgs the packages that might need to be installed
+   * @param {Object} consumerPackages the packages already installed in the consumer
+   * @returns {Array} the packages to install
+   */
+  getPkgsToInstall: function (pkgs, consumerPackages) {
+    const packagesVersionsPromise = this._getPackagesVersions(pkgs)
+    return packagesVersionsPromise.then((results) => {
+      const toInstall = []
+      results.forEach(({pkg, result}) => {
```
sglanzer-fire, line 18: Could run this in a `reduce` (no need to change, just noting)

package.json
```diff
@@ -10,8 +10,9 @@
     "build": "ember build",
     "lint": "lint-all-the-things",
     "start": "ember server",
-    "test": "npm run lint && npm run test-addon",
-    "test-addon": "EMBER_TRY_SCENARIO=${EMBER_TRY_SCENARIO:=default} && ember try:one $EMBER_TRY_SCENARIO --- COVERAGE=true ember test"
+    "test": "npm run lint && npm run test-addon && npm run test-cli",
+    "test-addon": "EMBER_TRY_SCENARIO=${EMBER_TRY_SCENARIO:=default} && ember try:one $EMBER_TRY_SCENARIO --- COVERAGE=true ember test",
+    "test-cli": "_mocha --recursive tests/unit/lib/"
```
sglanzer-fire: What does the underscore on mocha do?

tests/unit/lib/utils/npm.js
```diff
@@ -0,0 +1,125 @@
+const expect = require('chai').expect
+const sinon = require('sinon')
+const childProcessPromise = require('child-process-promise')
+const Promise = require('promise')
+
+const npmModuleLocation = '../../../../lib/utils/npm'
```
sglanzer-fire, line 6: Would an absolute path be easier here?

sglanzer-fire:
I mainly care about the null guard and the _mocha comments, the rest are optional - nice work, I'm excited to see the improvements this brings

lib/utils/npm.js
dafortin: We can run a command in `ember-cli` but we are not able to catch the output of the command. In the `view` case we need the output of the command.

tests/unit/lib/utils/npm.js
dafortin: I was having trouble with the absolute path for some obscure reason :(

sglanzer-fire: No problem, just curious

package.json
dafortin: https://github.com/gotwarlost/istanbul/issues/436

I changed it to mocha since we don't use istanbul.

blueprints/ember-frost-core/index.js
```diff
@@ -1,35 +1,33 @@
...
-            {name: 'ember-concurrency', target: '~0.7.15'},
-            {name: 'ember-computed-decorators', target: '~0.2.0'},
-            {name: 'ember-elsewhere', target: '~0.4.1'},
-            {name: 'ember-frost-test', target: '^1.0.0'},
-            {name: 'ember-prop-types', target: '^3.0.0'},
-            {name: 'ember-truth-helpers', target: '^1.2.0'}
-          ]
+          packages: pkgsToInstall,
+          blueprintOptions: {
+            saveDev: true
```
dafortin, line 26: In `ember-cli@2.8` they remove the default behavior so we need to specify this.
https://github.com/ember-cli/ember-cli/issues/6318

They put the default behavior back in `ember-cli@2.11`

dafortin:
Handled all the PR comments.

sglanzer-fire (approved):

sglanzer-fire:
@dafortin you mentioned you're hitting an issue with 2.8?  I assume that's blocking this?

dafortin:
No there is a workaround this issue: https://github.com/ciena-frost/ember-frost-core/pull/423/files#diff-f30e0028ce0a891ada9826aad7e7ade1R26

PR #425, opened 2017-03-29
added 'width' property to 'frost-select' interface

cstolli:

# CHANGELOG
* adds `width` property to the `frost-select` component interface.  works by explicitly overriding any width/min/max width on that `frost-select` instance to explicitly set a desired width.

### DIFF
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -237,7 +237,7 @@ export default Component.extend({
   /**
    * Get a style string based on the presence of some properties
    * @param {String} width the width property specified
-   * @return {String} the completed style string
+   * @returns {String} the completed style string
    */
   style (width) {
     let styles = ''
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -54,7 +54,8 @@ export default Component.extend({
     'ariaLabel:aria-label',
     'computedTabIndex:tabIndex',
     'opened:aria-pressed',
-    'role'
+    'role',
+    'style'
   ],
 
   classNameBindings: [
@@ -95,6 +96,7 @@ export default Component.extend({
       PropTypes.string
     ]),
     tabIndex: PropTypes.number,
+    width: PropTypes.oneOfType([PropTypes.number, PropTypes.null]),
     wrapLabels: PropTypes.bool,
 
     // state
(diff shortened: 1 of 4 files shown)


agonza40:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/425/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-select.js
```diff
@@ -228,6 +230,24 @@ export default Component.extend({
...
+   * @param {String} width the width property specified
+   * @returns {String} the completed style string
+   */
+  style (width) {
+    let styles = ''
+
+    // if a property is not falsy, append it to the style string
+    // note that in the width case, we want the component interface to have absolute power over the width
+    // so it will override any max or win widths to ensure ultimate control
+    if (width) styles += `width: ${width}px; max-width: initial; min-width:initial; `
```
agonza40, line 248: I thought we had a linting rule against if-statements without blocks. Also, why are we appending to styles if we know styles is empty otherwise? This could just be an assignment.

cstolli: there is no linting rule against it.

I figured there will be more styles that get implemented, and this way the line just works regardless of order of additional styles.

agonza40:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/425/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
I've only glanced so excuse the question - but why not handle this via CSS?

cstolli:
@sglanzer a programmatic approach was desirable so that it could be exposed as a bunsen/uis2 option, giving the form designer ultimate control when they know the data requires it.

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/425/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/components/frost-select.js
sandersky: If you're aiming for future support I'd have done:
```
style (width) {
  props = {}

  if (width) {
    assign(props, {
      'max-width': 'initial',
      'min-width': 'initial',
      width: `${width}px`
    })
  }

  // Convert props object to CSS string for styles attribute
  return Object.entries(props)
    .map(([key, value]) => `${key}: ${value}`)
    .join(';')
}
```

PR #427, opened 2017-03-31
feature(icon-fingerprinting): Scaffolding for iconAsset fp

### LINKED ISSUE #424: Icon packs are not fingerprinted
When doing a production build, most of our base assets are fingerprinted, which helps with cache issues when we update the app. However, icon packs are not among those resources which receive fingerprints.

The current implementation of `frost-icon` would not work if the icon packs received fingerprints either: https://github.com/ciena-frost/ember-frost-core/blob/81e9629a75d5b75ca090fcd454309e39f7c6f87d/addon/templates/components/frost-icon.hbs#L2

It would be great if we had some mechanism for identifying the version of the icon-pack `.svg` files so that browsers always had the proper version w/o users needing to clear their cache.

srowhani:

For now this intends to be a proof of concept, and should not be merged.

### How it works
- Use broccoli-asset-rev to wrap pre-existing `treeForPublic`
    - By default b-a-r will ignore svg files
- Generate asset map
- Create an application initializer that defers app readiness state and injects assetMap
- Frost icon refers to the icon pack through a CP

# CHANGELOG
- Icon packs are now fingerprinted

Closes #424

### DIFF
--- a/addon/components/frost-icon.js
+++ b/addon/components/frost-icon.js
@@ -6,14 +6,13 @@ import Component from './frost-component'
 import Ember from 'ember'
 import computed, {readOnly} from 'ember-computed-decorators'
 import {PropTypes} from 'ember-prop-types'
-const {deprecate, inject, get} = Ember
+const {deprecate, get} = Ember
 
 export default Component.extend({
 
   // == Dependencies ==========================================================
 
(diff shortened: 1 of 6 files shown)


srowhani:
Closed in favor of #429

PR #433, opened 2017-04-12
Ember 2.12 Upgrade

sandersky:

# CHANGELOG

* **Upgraded** to Ember 2.12.


sglanzer-fire:
Approved - just to clarify for anyone looking at this review - the LTS recommendation for Frost is still 2.8

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/433/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
Sorry - hold on this for a minute, a few questions incoming

sglanzer-fire (changes requested):
@juwara0 has a few tweaks that should go in with the 2.12 upgrade to align with the ember-cli blueprints.  I'd like to get them into this upgrade so that we don't forget

juwara0:
Lets make sure we get all of the updates: I noticed that the update from 2.11 to the resolvers are still not in...

https://github.com/ciena-frost/ember-frost-object-browser/pull/113/files#diff-5265b30549af2bba1fba0ab3b2165e54
https://github.com/ciena-frost/ember-frost-object-browser/pull/113/files#diff-2be9329279338d3d4f10bea51caa639e

sandersky:
It looks like we may be in for a bunch of pain given all pre-2.12 CI builds fail (I'm guessing because we switch from an ember bower dependency to the ember-source npm dependency)

bower.json
```diff
@@ -1,17 +1,10 @@
 {
   "name": "ember-frost-core",
   "dependencies": {
-    "ember": "2.11.2",
-    "ember-cli-shims": "0.1.3",
-    "ember-inflector": "^1.3.1",
-    "ember-mocha": "0.8.11",
+    "ember": "~2.12.1",
```
juwara0, line 4: This should not be here. It should be coming from `ember-source` npm package.

juwara0:
As a general comment not directed at @sandersky (thanks for doing this upgrade):

When upgrading please run `ember init` and look closely at the diffs. The small changes to paths are easy to miss.

juwara0 (changes requested):

sglanzer-fire:
@juwara0 I don't suppose we can script those changes?  I've done them before far too many times and it's painstaking to do the file by file compare even if it is necessary

sglanzer-fire:
...or at least say - these _x_ are fine to overwrite - pay attention to _y_

bower.json
sandersky: It actually does need to be here because for whatever reason we are including the template compiler https://github.com/ciena-frost/ember-frost-core/blob/master/ember-cli-build.js#L26. It looks like you were the one who added that import @juwara0

sandersky:
I ran `ember init` but it overwrites entire files rather than applying changes to them so it is quite possible I just missed the changes in the the process and thus didn't keep them. The fact `ember init` does this makes for some really lousy DX.

juwara0:
When running `ember-init` it will give you several choices. One of which is to see a diff. I select this option so I can see if the changes are just our removal of semicolons, moving order of imports, etc... or something that is needed. For the needed changes, I open the file and make those changes myself.

bower.json
juwara0: To be clear, the move of `ember` from `bower.json` to npm as `ember-source` was supposed to happen as part of the 2.11.x upgrade. It is also correct that the path to the template compiler will need to be updated.

Issue #281, opened 2016-11-16
Multi-attribute select


kparmeso:
We have use cases that require multiple attributes in order for a user to make a choice when using the select tool.

The user should be able to filter the options on any attribute, primary or secondary.
Secondary attributes should appear below the Primary separated by a | if there are more than one.

![select-multi-attribute](https://cloud.githubusercontent.com/assets/17346066/20352182/40b1f720-abcb-11e6-8fed-e63ba2ff8cf8.png)

PR #434, opened 2017-04-18
Adding multi labels to select

helrac:

# CHANGELOG
- Added secondary labels in select #281

### DIFF
--- a/addon/components/frost-select-dropdown.js
+++ b/addon/components/frost-select-dropdown.js
@@ -2,7 +2,7 @@
  * Component definition for frost-select-dropdown component
  */
 import Ember from 'ember'
-const {$, deprecate, get, merge} = Ember
+const {$, deprecate, get, isArray, isEmpty, merge} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import {task, timeout} from 'ember-concurrency'
 import {PropTypes} from 'ember-prop-types'
@@ -79,7 +79,6 @@ export default Component.extend({
   },
 
   // == Computed Properties ===================================================
-
   @readOnly
   @computed('wrapLabels')
   /**
@@ -106,6 +105,21 @@ export default Component.extend({
     const classNames = ['frost-select-list-item-text']
     if (multiSelect) {
       classNames.push('frost-multi-select-list-item-text')
+    }
+    return classNames.join(' ')
+  },
+
+  @readOnly
+  @computed('multiselect')
+  /**
+   * The class names for the frost select dropdown secondary options
+   * @param {Boolean} multiSelect - whether or not this is a multiselect
+   * @returns {string} the class names for the frost select dropdown secondary options
+   */
+  dropdownSecondaryLabelsTextClassNames (multiSelect) {
+    const classNames = ['frost-select-list-secondary-item-text']
+    if (multiSelect) {
+      classNames.push('frost-multi-select-list-secondary-item-text')
     }
     return classNames.join(' ')
   },
@@ -184,6 +198,11 @@ export default Component.extend({
       const classNames = ['frost-select-list-item']
       const value = get(item, 'value')
       const isSelected = selectedItems.find((item) => item.value === value) !== undefined
+      const secondaryLabels = get(item, 'secondaryLabels')
+
+      if (!isEmpty(secondaryLabels)) {
+        classNames.push('frost-select-list-secondary-item')
+      }
 
       if (isSelected) {
         classNames.push('frost-select-list-item-selected')
@@ -196,6 +215,25 @@ export default Component.extend({
       return {
         className: classNames.join(' '),
         label: get(item, 'label'),
+        secondaryLabels: get(item, 'secondaryLabels'),
+
+        @readOnly
+        @computed('secondaryLabels')
+        hasSecondaryLabels (secondaryLabels) {
+          if (isArray(secondaryLabels)) {
+            return true
+          }
+          return false
+        },
+
+        @readOnly
+        @computed('secondaryLabels')
+        displaySecondaryLabels (secondaryLabels) {
+          if (isArray(secondaryLabels)) {
+            return secondaryLabels.join(' | ')
+          }
+          return ''
+        },
         selected: isSelected,
         value: get(item, 'value')
       }
@@ -372,6 +410,9 @@ export default Component.extend({
     const clonedDropdownListElement = dropdownListElement.cloneNode(true)
     const clonedTextElements = clonedDropdownListElement.querySelectorAll('.frost-select-list-item-text')
     const textElements = dropdownListElement.querySelectorAll('.frost-select-list-item-text')
+    const clonedSecondaryTextElements = clonedDropdownListElement
+                                        .querySelectorAll('.frost-select-list-secondary-item-text')
+    const secondaryTextElements = dropdownListElement.querySelectorAll('.frost-select-list-secondary-item-text')
     const scrollTop = dropdownListElement.scrollTop
     const wrapLabels = this.get('wrapLabels')
 
@@ -401,6 +442,27 @@ export default Component.extend({
       }
     })
 
+    Array.from(secondaryTextElements).forEach((secondaryTextElement, index) => {
+      if (!wrapLabels) {
+        const clonedSecondaryTextElement = clonedSecondaryTextElements[index]
+        const updatedData = trimLongDataInElement(clonedSecondaryTextElement)
+
+        if (updatedData) {
+          secondaryTextElement.textContent = updatedData.text
+          secondaryTextElement.setAttribute('title', updatedData.tooltip)
+        }
+      }
+
+      if (filter) {
+        const pattern = new RegExp(filter, 'gi')
+        const textWithMatch = secondaryTextElement.textContent.replace(pattern, '<u>$&</u>')
+
+        // If rendered text has changed, update it
+        if (secondaryTextElement.innerHTML !== textWithMatch) {
+          secondaryTextElement.innerHTML = textWithMatch
+        }
+      }
+    })
     clonedDropdownListElement.replaceWith(dropdownListElement)
 
     this._addListItemEventListeners(dropdownListElement)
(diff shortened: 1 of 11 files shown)


helrac:
image

dafortin:
Is the spacing between the primary and secondary label following the UX specs? The primary label is very close to the top of the `item`.

addon/components/frost-select-dropdown.js
```diff
@@ -111,6 +110,21 @@ export default Component.extend({
   },
 
   @readOnly
+  @computed('multiselect')
+  /**
+   * The class names for the frost select dropdown secondary options
+   * @param {Boolean} multiSelect - whether or not this is a multiselect
```
dafortin, line 116: Any reason why `Boolean` is in caps and `string` in small?

addon/components/frost-select-dropdown.js
```diff
@@ -196,6 +215,25 @@ export default Component.extend({
       return {
         className: classNames.join(' '),
         label: get(item, 'label'),
+        secondaryLabels: get(item, 'secondaryLabels'),
+
+        @readOnly
+        @computed('secondaryLabels')
+        hasSecondaryLabels (secondaryLabels) {
```
dafortin: Is there is specific reason to use a cp here instead of using simply…
`hasSecondayLabels: isArray(secondaryLabels)`

addon/components/frost-select-dropdown.js
```diff
@@ -196,6 +215,25 @@ export default Component.extend({
...
+          if (isArray(secondaryLabels)) {
+            return true
+          }
+          return false
+        },
+
+        @readOnly
+        @computed('secondaryLabels')
+        displaySecondaryLabels (secondaryLabels) {
```
dafortin: same comment as above

addon/components/frost-select-dropdown.js
```diff
@@ -401,6 +442,27 @@ export default Component.extend({
       }
     })
 
+    Array.from(secondaryTextElements).forEach((secondaryTextElement, index) => {
```
dafortin: This part of the code looks a lot like line 382. Is there a way to avoid duplication?

addon/components/frost-select.js
```diff
@@ -169,8 +170,18 @@ export default Component.extend({
...
         const label = item.label || ''
+        const secondaryLabels = item.secondaryLabels || []
 
-        return label.toLowerCase().indexOf(filter) !== -1
+        if (label.toLowerCase().indexOf(filter) !== -1) {
+          return true
+        }
+
+        const filteredSecondaryLabels = secondaryLabels.filter(function (secondaryLabel) {
+          if (secondaryLabel.toLowerCase().indexOf(filter) !== -1) {
```
dafortin: I think we can directly do `return scondaryLabel.... != -1`

addon/styles/_frost-select.scss
```diff
@@ -203,6 +204,47 @@ $frost-select-list-item-horizontal-padding: 10px;
...
+  .frost-select-list-secondary-item-text, .frost-multi-select-list-secondary-item-text {
+    line-height: $frost-font-s;
+    min-height: $frost-font-s;
+    font-size: $frost-font-s;
+    padding-top: 5px;
+    padding-bottom: 3px;
+    color: $frost-color-grey-5;
+    font-weight: 100;
+
+    u {
```
dafortin: Maybe we should use a class instead of the `u` element?

addon/templates/components/frost-select-dropdown.hbs
```diff
@@ -63,12 +63,22 @@
             size='medium'
           }}
         {{/if}}
-        <span
+        <div class="frost-select-list-item-container">
```
dafortin: Might be a github thing but looks like the indentation is off between this div and the div below

addon/components/frost-select-dropdown.js
helrac: I used the same formatting as dropdownTextClassNames

dafortin: I think they are both fine, but we should try to be consistent. 

JSDoc doesn't really care :) https://github.com/jsdoc3/jsdoc/issues/1046

addon/components/frost-select-dropdown.js
helrac: Was just a byproduct of me changing things around. I'll update it

dafortin: :)

addon/templates/components/frost-select-dropdown.hbs
helrac: It was not a github thing, the nested indentation was just off

dafortin: ok great :)

sandersky:
Why are you fixing toggle and select in a single PR? Generally it is a good idea to solve multiple bugs in separate PR's, especially when the bugs are on completely different components.

package.json
```diff
@@ -88,6 +88,7 @@
     "broccoli-funnel": "^1.0.1",
     "broccoli-merge-trees": "^1.0.0",
     "broccoli-svgstore": "git://github.com/ciena-frost/broccoli-svgstore.git",
+    "ember-cli": "^2.11.1",
```
sandersky: I don't think this should have been added.

addon/templates/components/frost-select-dropdown.hbs
```diff
@@ -63,12 +63,22 @@
...
           {{item.label}}
-        </span>
+        </div>
+        {{#if item.hasSecondaryLabels}}
+        <div
+          class={{dropdownSecondaryLabelsTextClassNames}}
+          data-text={{item.displaySecondaryLabels}}
+        >
+          {{item.displaySecondaryLabels}}
+        </div>
```
sandersky: Lines 74-79 should be tabbed (2 spaces) to the right for proper indentation.

addon/styles/_frost-toggle.scss
```diff
@@ -61,6 +61,13 @@ $frost-toggle-on-color: #fff;
       content: '';
     }
 
+    @-moz-document url-prefix() {
```
sandersky: Can you add a comment here as to what exactly this is doing/achieving?

sandersky:
I've also noticed you didn't add any tests to exercise the bug you are fixing nor the new feature add to the select. It is pertinent we add tests for every PR as to keep things from regressing in the future.

addon/components/frost-select-dropdown.js
```diff
@@ -196,6 +215,25 @@ export default Component.extend({
       return {
         className: classNames.join(' '),
         label: get(item, 'label'),
+        secondaryLabels: get(item, 'secondaryLabels'),
```
sandersky: You already looked this property up on line 201. Instead this line can simply be:
```
secondaryLabels,
```

addon/components/frost-select-dropdown.js
```diff
@@ -196,6 +215,25 @@ export default Component.extend({
...
+        },
+
+        @readOnly
+        @computed('secondaryLabels')
+        displaySecondaryLabels (secondaryLabels) {
+          if (isArray(secondaryLabels)) {
+            return secondaryLabels.join(' | ')
+          }
+          return ''
+        },
```
sandersky: This is a great place to use the ternary operator.
```
displaySecondaryLabels (secondaryLabels) {
  return isArray(secondaryLabels) ? secondaryLabels.join(' | ') : ''
},
```

addon/components/frost-select-dropdown.js
```diff
@@ -372,6 +410,9 @@ export default Component.extend({
     const clonedDropdownListElement = dropdownListElement.cloneNode(true)
     const clonedTextElements = clonedDropdownListElement.querySelectorAll('.frost-select-list-item-text')
     const textElements = dropdownListElement.querySelectorAll('.frost-select-list-item-text')
+    const clonedSecondaryTextElements = clonedDropdownListElement
+                                        .querySelectorAll('.frost-select-list-secondary-item-text')
```
sandersky: This mega tabbing isn't how we usually do it. Instead I'd go with:
```
const clonedSecondaryTextElements = clonedDropdownListElement
  .querySelectorAll('.frost-select-list-secondary-item-text')
```
Otherwise if anyone ever renames the variable in line 413 they'll also need to adjust the spacing in line 414.

package.json
helrac: I'll remove it

sglanzer-fire:
Tests are in progress for this afaik - @helrac ping me when it's ready for a re-review

@sandersky I OKed the inclusion the toggle fix, just easier than jiggling git around at the time

addon/templates/components/frost-select-dropdown.hbs
```diff
@@ -63,22 +63,31 @@
...
+            data-test={{hook (concat hookPrefix '-secondaryLabel') index=index}}
           >
             {{item.displaySecondaryLabels}}
           </div>
-        {{/if}}
         </div>
+        {{else}}
+        <div
+          class={{dropdownTextClassNames}}
+          data-text={{item.label}}
```
sglanzer-fire: Might as well through a `data-test` hook on here as well - `text()` will still function, but consistency would be a positive move

tests/integration/components/frost-select-test.js
```diff
@@ -562,6 +574,197 @@ describe(test.label, function () {
...
+        it('renders as expected', function () {
+          expectSelectWithState('select', {
+            focused: true,
+            focusedItem: 'Foo',
+            items: ['Foo', 'Bar', 'Ba ba'],
+            secondaryLabels: ['Foo1-1 | Foo1-2', 'Bar1-1 | Bar1-2', 'Ba | Ba | Black Sheep'],
+            opened: true
+          })
+        })
+      })
```
sglanzer-fire: Super minor, but I'd add a line between describe blocks

sglanzer-fire:
Two really minor comments to address

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/434/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #438, opened 2017-04-26
Fix frost toggle click handler

job13er:

Since I made a few, unrelated changes in this PR, I tried to group my commits in a meaningful way. By looking at the diffs per commit, you should be able to understand what changes were needed for what effect. If it's unclear, just ask. 

# CHANGELOG

* **Upgraded** to `ember-test-utils@4.x`
* **Upgraded** to `ember-cli-code-coverage@0.3.12` (using a pinned `ember-code-snippet@1.7.0` so as to avoid overwriting `require` and breaking `ember-cli-code-coverage`)
* **Enabled** code coverage gating in PRs. 
* **Added** `onToggle` callback which is called with the new value whenever toggled
* **Deprecated** `onClick` callback, which wasn't handling raw functions very well when going through the event proxy
* **Fixed** [#439](https://github.com/ciena-frost/ember-frost-core/issues/439) Properly handle falsy `trueValue`/`falseValue` when `trueLabel`/`falseLabel` are set


.eslintrc.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
```
job13er, line 1: No actual change in this guy, just reformatting from `.eslintrc` to `.eslintrc.js` b/c javascript > json :wink:

package.json
```diff
@@ -46,7 +46,7 @@
     "ember-cli-notifications": "^4.1.6",
     "ember-cli-test-loader": "^1.1.1",
     "ember-cli-uglify": "^1.2.0",
-    "ember-code-snippet": "^1.8.0",
+    "ember-code-snippet": "ciena-blueplanet/ember-code-snippet#remove-browserify",
```
job13er: I did raise an [issue](https://github.com/ef4/ember-code-snippet/issues/34) with `ember-code-snippet` to make them aware of the issue, but in the meantime, we can just use our fork.

package.json
```diff
@@ -98,5 +98,8 @@
     "ember-spread": "^1.1.1",
     "fs-extra-promise": "^0.4.0",
     "npm-install-security-check": "^1.0.0"
+  },
+  "pr-bumper": {
+    "coverage": 92.72
```
job13er: This will help ensure that code coverage doesn't drop as the result of a PR

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/438/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
Looks like a few tests need to be updated to deal with true/false CPs in the toggle being readOnly - I'd probably recommend removing those unit tests since we've agreed that integration tests should cover components in a black-box approach

job13er:
> Looks like a few tests need to be updated to deal with true/false CPs in the toggle being readOnly - I'd probably recommend removing those unit tests since we've agreed that integration tests should cover components in a black-box approach

Sounds good. I was gonna refactor the unit tests, but just updating the integration tests to make sure all the use-cases are covered sounds much better.

job13er:
@sglanzer Worth taking another look, this is a bit different since you approved last. It now implements the `onToggle` as we discussed in slack, and deprecates `onClick`

job13er:
Hold on, just found another bug, closing this while I fix it. Will re-open once I have the new tests/fix in place.

job13er:
OK @sglanzer I think it's ready for review again. Sorry for the hold-up.

job13er:
Hmm.

>pr-bumper: ERROR: Code Coverage: `91.94%` (dropped `0.78%` from `92.72%`)

Looks like I need to add a couple more tests. Coming right up.

addon/components/frost-toggle.js
```diff
@@ -53,45 +58,44 @@ export default Component.extend(SpreadMixin, PropTypeMixin, FrostEventsProxyMixi
   },
 
   getDefaultProps () {
+    const falseLabel = this.get('falseLabel')
+    const trueLabel = this.get('trueLabel')
+    const validTypes = ['string', 'number', 'boolean']
```
sglanzer-fire, line 63: I'm honestly not sure why we would care about the type of the label - I know this was there in the previous code, I just can't think of any reason for it

job13er: Me either. The label should always be a string, and spec'd that way in propTypes in my opinion, but I didn't want to change functionality in this PR, just clean up the code :)

addon/components/frost-toggle.js
```diff
@@ -53,45 +58,44 @@ export default Component.extend(SpreadMixin, PropTypeMixin, FrostEventsProxyMixi
...
   // == Computed Properties ===================================================
 
+  @readOnly
   @computed('trueValue', '_trueLabel')
-  // FIXME: jsdoc
-  // TODO: make computed property readOnly
-  _trueValue (trueValue, _trueLabel) { // eslint-disable-line
-    return trueValue || _trueLabel
+  _trueValue (trueValue, _trueLabel) {
+    return (trueValue === undefined) ? _trueLabel : trueValue
```
sglanzer-fire, line 82: Probably another case where we can't change this due to backwards compatibility, but it probably makes more sense to default to `true` instead of using the label as the value (typically we would use the value as a label, not the other way around)

addon/components/frost-toggle.js
```diff
@@ -53,45 +58,44 @@ export default Component.extend(SpreadMixin, PropTypeMixin, FrostEventsProxyMixi
...
   },
 
+  @readOnly
   @computed('falseValue', '_falseLabel')
-  // FIXME: jsdoc
-  // TODO: make computed property readOnly
-  _falseValue (falseValue, _falseLabel) { // eslint-disable-line
-    return falseValue || _falseLabel
+  _falseValue (falseValue, _falseLabel) {
+    return (falseValue === undefined) ? _falseLabel : falseValue
```
sglanzer-fire, line 88: Same obviously - likely we can't do anything about these comments, but maybe we can make an LTS issue pointing to this PR so that we can cleanup in the next release

sglanzer-fire:
Approved - all good from my end - re-kicked a few failed builds (3 passed, it's probably just travis)

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/438/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

job13er:
Could be the code coverage still. I'll take a look and either adjust the baseline (since I'm only enabling it in this PR anyway) or add some missing tests elsewhere to increase coverage. I did make sure I was covering toggle well. The only bit not covered was some code I couldn't figure out how to reach, but didn't want to delete for backwards compatibility.

sglanzer-fire:
Raised an issue to track the comments we can't address currently https://github.com/ciena-frost/ember-frost-core/issues/440

sglanzer-fire:
@job13er yeah, you're getting code coverage drops, but only in 2 of the builds (used to be 3) - we may need to be careful about that strategy

job13er:
Yeah, should only be checking in one build, I'll fix it.

juwara0:
@job13er - As a reference, here is what I did to limit the checking to one build:
https://github.com/ciena-frost/ember-frost-object-browser/pull/115/files#diff-354f30a63fb0907d4ad57269548329e3
https://github.com/ciena-frost/ember-frost-object-browser/pull/115/files#diff-b29511921f39def9f1eec87ca24e0e4d

job13er:
Something screwy is going on. I shouldn't be getting different coverage numbers for different builds. We don't have conditional code based on ember/noe version (to the best of my knowledge)

sglanzer-fire:
Yeah, I'm not sure what that's about - maybe we only want to run coverage numbers against the LTS release build?

.travis/maybe-check-coverage.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
job13er, line 1: This should let us only check coverage on the same build where we bump.

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/438/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

job13er:
I'm stuck behind bunsen builds, so it may be a while :(

job13er:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/438/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

sglanzer-fire:
Woooh, finally got it merged - good stuff 👍

PR #444, opened 2017-05-10
Added onClick callback for frost-select

TheOtherDude:

This callback might be useful if you want to `stopPropagation` of the click event.

# CHANGELOG

* **Added** onClick callback for frost-select

### DIFF
--- a/addon/components/frost-select.js
+++ b/addon/components/frost-select.js
@@ -275,9 +275,12 @@ export default Component.extend({
 
   // == DOM Events ============================================================
 
-  _onClick: on('click', function () {
+  _onClick: on('click', function (e) {
     if (!this.get('disabled')) {
       this.toggleProperty('opened')
+    }
+    if (this.onClick) {
+      this.onClick(e)
     }
   }),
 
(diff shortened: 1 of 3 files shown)


gknoy:
👍

sglanzer-fire:
Can you add a description to the PR of the use case being solved?

TheOtherDude:
Description updated.

job13er:
@TheOtherDude If it didn't previously have an `onClick`, wouldn't this be a MINOR change, adding functionality, not a patch bugfix? A little late now, since it was already merged/released, but for future...

TheOtherDude:
@job13er Yeah, that would have made more sense. My bad!

sglanzer-fire:
Late to the party here (email backlog kills me) - for the future when I'm asking for the use case I'm looking for the root problem that you're trying to solve, not the capability that the change provides.  It's much easier for me to decide if a change should be made if I understand _why_ the change is being made and in this case I'm still not sure.

This doesn't just apply to this PR, but in general our PR descriptions are quite poor in describing the intention and use case for the PR.  If you could provide me with the use case for this I would appreciate it - I don't really care that it's post-merge, but I still want to know.

PR #445, opened 2017-05-11
Export windowUtils as part of utils and add frost-link test helper

job13er:

# CHANGELOG

* **Added** missing export of `windowUtils` inside `utils` that's exported from `addon/index.js`. Consumers can now import the `windowUtils` without needing to know the   path in the filesystem:
  ```js
  import {utils} from 'ember-frost-core'
  const {windowUtils} = utils
  ``` 
  instead of
  ```js
  import windowUtils from 'ember-frost-core/utils/window'
  ```
 * **Added** `frost-link` test helpers to `addon-test-support`, specifically just a `stubRouteService` helper method to allow consumers of `frost-link` to check the routes passed into it. 
* **Cleaned** up the integration tests for `frost-link` a little.

### DIFF
--- a/addon-test-support/frost-link.js
+++ b/addon-test-support/frost-link.js
@@ -23,9 +23,7 @@ export function stubRoutingService (context) {
       return true
     },
 
-    transitionTo () {
-      //
-    }
+    transitionTo () {}
   })
 
   context.registry.register('service:-routing', RouterStub)
--- a/addon-test-support/frost-link.js
+++ b/addon-test-support/frost-link.js
@@ -1,0 +1,32 @@
+/**
+ * Test helpers for frost-link
+ */
+import Ember from 'ember'
+
+/**
+ * Stub the routing service so we can test the route given to a frost-link
+ * @param {*} context - the testing context (`this` from within your beforeEach())
+ */
+export function stubRoutingService (context) {
+  const RouterStub = Ember.Object.extend({
+    generateURL (endpoint, segments) {
+      let url = endpoint
+
+      if (segments.length !== 0) {
+        url = `${url}/${segments.join('/')}`
+      }
+
+      return url
+    },
+
+    hasRoute () {
+      return true
+    },
+
+    transitionTo () {
+      //
+    }
+  })
+
+  context.registry.register('service:-routing', RouterStub)
+}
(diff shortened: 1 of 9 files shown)


addon-test-support/frost-link.js
```diff
@@ -0,0 +1,32 @@
...
+
+      return url
+    },
+
+    hasRoute () {
+      return true
+    },
+
+    transitionTo () {
+      //
```
dafortin: Do we need this?

addon-test-support/frost-link.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Stub the routing service so we can test the route given to a frost-link
+ * @param {*} context - the testing context (`this` from within your beforeEach())
+ */
+export function stubRoutingService (context) {
+  const RouterStub = Ember.Object.extend({
+    generateURL (endpoint, segments) {
+      let url = endpoint
+
+      if (segments.length !== 0) {
```
dafortin, line 15: Any reason to use this instead of `isEmpty`?

addon-test-support/frost-link.js
job13er: I don' know, I only moved the content that already existed from the test of `frost-link` to the `addon-test-support/` so others can use it.

dafortin: Ah ok can we try to remove `transitionTo` and if it's necessary add a comment or something like that.

addon-test-support/frost-link.js
job13er: Again, not my code, I just moved it, but I'd only use `isEmpty` if I needed to handle multiple types of data, or multiple possible "empty" values. In this case, `segments` is known to be an array, and so an explicit check on the length is, IMO, clearer and more appropriate.

dafortin: ok 👍

addon-test-support/frost-link.js
job13er: Sure, seems like that's something that should have been addressed in the original PR, and not the one that is just moving things around, but ok.

dafortin (approved):
Looks good :)

job13er:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/445/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

Issue #405, opened 2017-02-17
Provide a numeric input component


sglanzer-fire:
Handle specific use cases such as increment/decrement and min/max, don't accept non-numeric input, work with decimals, potentially provide units, etc.

sglanzer-fire:
![image](https://cloud.githubusercontent.com/assets/9026198/23949441/9095ea86-095d-11e7-90db-647e1ef1b4f4.png)

sglanzer-fire:
![image](https://cloud.githubusercontent.com/assets/9026198/23949452/9f00bce0-095d-11e7-8dd1-db79984cc0f4.png)

notmessenger:
I would like to see all of the HTML attributes fully supported, such as how are broken out here - https://github.com/ember-forge/ember-forge-ui/blob/form-components/addon/components/form/input/ef-number.js

PR #449, opened 2017-05-17
new Numeric field component

helloWorldCindy:

# CHANGELOG

Add numeric selection component


sglanzer-fire:
So before I start the review I just wanted to give you a heads up - don't be worried if there is a volume of comments - one of Carl's latest reviews had around 90+, it's simply and educational thing and we tend to be picky since we're productizing components for multiple apps to consume.

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
+/**
+ * Component definition for the numeric selection field component
+ */
+import layout from '../templates/components/frost-numeric-select'
+import Component from './frost-component'
+import Ember from 'ember'
+import {PropTypes} from 'ember-prop-types'
+import computed, {readOnly} from 'ember-computed-decorators'
+const {isPresent} = Ember
```
sglanzer-fire: We've been moving towards a convention recently of grouping external (from other packages) and internal (from the same package) imports - which would look like

```
import Ember from 'ember'
const {isPresent} = Ember
import computed, {readOnly} from 'ember-computed-decorators'
import {PropTypes} from 'ember-prop-types'

import Component from './frost-component'
import layout from '../templates/components/frost-numeric-select' 
```

Also note the alphabetical ordering and putting the destructuring of an import directly beside the import

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
+/**
+ * Component definition for the numeric selection field component
```
sglanzer-fire: I'm going to suggest naming this component something other than a `-select` since there is a very specific connotation for that word.  How about `frost-numeric-field`?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+  // == PropTypes =============================================================
+
+  propTypes: {    // options
+    align: PropTypes.string,
+    isHookEmbedded: PropTypes.bool,
+    receivedHook: PropTypes.string,
+    tabindex: PropTypes.number,
+    type: PropTypes.string,
+    isArrowVisible: PropTypes.bool,
+    isAllowInputNegative: PropTypes.bool,
```
sglanzer-fire: How about `allowNegativeValues` instead of the slightly awkward `isAllow`?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+    isAllowUseDownButtonToNegative: PropTypes.bool,
```
sglanzer-fire: Can we calculate this off the `isAllowInputNegative`?  Seems like it would be possible to remove this from the interface

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+    autocapitalize: PropTypes.string,
+    autofocus: PropTypes.bool,
+    autocorrect: PropTypes.string,
+    form: PropTypes.string,
+    maxlength: PropTypes.number,
+    placeholder: PropTypes.string,
+    required: PropTypes.bool,
+    selectionDirection: PropTypes.string,
+    spellcheck: PropTypes.bool,
+    value: PropTypes.string,
```
sglanzer-fire: I believe the `value` should be required

sglanzer-fire: Ember (and other frameworks) promote a concept known in Ember as Data Down, Actions Up https://dockyard.com/blog/2016/11/18/checkbox-list-ember

Basically the idea is that the state required for a component to function should be held outside of the component and component events should fire to that layer to mutate the state.

This means that `value` is your Data Down and you need a callback like `onInput` to act as the Action Up.

A consumer would have a property that would store the actual state, something like 

https://github.com/ciena-frost/ember-frost-core/blob/master/tests/dummy/app/pods/toggle/template.hbs#L12
and
https://github.com/ciena-frost/ember-frost-core/blob/master/tests/dummy/app/pods/toggle/controller.js#L23

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+      // options
+      align: 'left',
+      autocapitalize: 'off',
+      autocorrect: 'off',
+      autofocus: false,
+      isArrowVisible: false,
+      isAllowInputNegative: true,
+      isAllowUseDownButtonToNegative:false,
+      showError: false,
+      errorMessage: 'Cannot input invalid character',
```
sglanzer-fire: Maybe `Input must be numeric`?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+      required: false,
+      spellcheck: false,
+      tabindex: 0,
+      type: 'text'
+    }
+  },
+
+  // == Functions =============================================================
+
+  _isNumberKey (e) {
```
sglanzer-fire: `core` has a keycode enum that would be good to use here https://github.com/ciena-frost/ember-frost-core/blob/master/addon/utils/key-codes.js

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+    let decimalExist = this.get('value') && this.get('value').indexOf('.') >=0
+    let charCode = (e.which) ? e.which : e.keyCode
+
+    if(!this.get('isAllowInputNegative')){
```
sglanzer-fire: Minor, but for clarity I would prefer to check the positive condition (without `!`) and put the negative condition in the `else`

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+
+  // == Computed Properties ===================================================
+  /**
+   * Determine which scss class should be used
+   * @param {boolean} isArrowVisible - if the user allow displaying the up and down arrow buttons
+   * @returns {string} - a scss class name
+   */
+  @computed('isArrowVisible')
+  class (isArrowVisible) {
+    return isArrowVisible ? 'frost-numeric-select-button' : 'frost-numeric-select'
```
sglanzer-fire: I'd like this to be a conditional modifier on the base class, so something like `.frost-numeric-select` as the base class and `frost-numeric-select.arrow-visible` as the modifier, where the `arrow-visible` would provide the right padding, etc.

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+
+  // == PropTypes =============================================================
+
+  propTypes: {    // options
+    align: PropTypes.string,
+    isHookEmbedded: PropTypes.bool,
+    receivedHook: PropTypes.string,
+    tabindex: PropTypes.number,
+    type: PropTypes.string,
+    isArrowVisible: PropTypes.bool,
```
sglanzer-fire: How about `isIncrementControlVisible`?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+      }
+    },
+
+    _onKeyPress (event) {
+      if (!this._isNumberKey(event)) {
+          this.set('showError', true)
+          event.preventDefault();
+      }
+      else {
+        this.set('showError', false)
```
sglanzer-fire: I don't think we can clear off the error if an invalid character is still present.  How about calculating the numeric error state using a computed property based on the `value`?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+      }
+    },
+
+    _onInput (event) {
+      if (isPresent(this.get('_eventProxy.input'))) {
+        // Add id and value for legacy support
```
sglanzer-fire: We don't need the legacy support here, so we can just send the event as-is

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+        event.id = this.get('elementId')
+        event.value = event.target.value
+        this._eventProxy.input(event)
+      }
+    },
```
sglanzer-fire: Really minor, but a linebreak is good between properties (and header comments)

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+    _onPlusClick (event) {
+      if (this.$('input').val()){
```
sglanzer-fire: Do we need the `$('input').val()` here?  Wouldn't `value` be a better choice?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+        //calculate how many digit a value has
+        let numberOfDigit = this._calculateDigit(this.$('input').val())
+        var value = parseFloat(this.get('value'))+1
+
+        if (!this.get('isAllowUseDownButtonToNegative')) {
+          if(value >= 0)
+            this._setCalculatedValue(value,numberOfDigit)
```
sglanzer-fire: This is where I would replace the `set` with an `onInput` callback

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+        this._eventProxy.input(event)
+      }
+    },
+    _onPlusClick (event) {
+      if (this.$('input').val()){
+        //calculate how many digit a value has
+        let numberOfDigit = this._calculateDigit(this.$('input').val())
+        var value = parseFloat(this.get('value'))+1
+
+        if (!this.get('isAllowUseDownButtonToNegative')) {
```
sglanzer-fire: Shouldn't be necessary for the `plus` right?

addon/styles/_frost-numeric-select.scss
```diff
@@ -0,0 +1,99 @@
+$height: 30px;
+$border: 0.3px solid $frost-color-input-border;
+$focus-border: 1px solid $frost-color-blue-1 !global;
+$frost-color-button-hover-bg: rgba($frost-color-blue-1, .1) !global;
+$arrow-focus-color: $frost-color-blue-1 !global;
+$error-text-color: $frost-color-input-error-border;
+$disabled-border: 1px solid $frost-color-input-disabled-border;
+$disabled-bg: $frost-color-input-disabled-bg;
```
sglanzer-fire: I'd like to see these variables prefixed since they're going to end up in the global namespace - so

`$frost-number-select-height: 30px;` etc.

addon/styles/_frost-numeric-select.scss
```diff
@@ -0,0 +1,99 @@
...
+  height: $height;
+  margin-bottom: 15px;
+  outline: none;
+  padding-right: 26px;
+
+  &:focus {
+    border: $focus-border;
+  }
+}
+.error {
```
sglanzer-fire: Be careful with generic classes - CSS is a global namespace so this could inadvertently apply to elements you wouldn't have expect to impact.  It's much safer to make this `frost-numeric-select-error`

addon/styles/_frost-numeric-select.scss
```diff
@@ -0,0 +1,99 @@
...
+  width: 175px;
+  height: $height;
+  font-size: 13px;
+  color: $error-text-color;
+}
+.container{
```
sglanzer-fire: Prefix

addon/styles/_frost-numeric-select.scss
```diff
@@ -0,0 +1,99 @@
...
+  border: $border;
+  border-right: none;
+  border-top: none;
+  border-bottom: none;
+  flex-direction: column;
+  position: absolute;
+  right: 0px;
+  top: 0.7px;
+}
+.minusButton {
```
sglanzer-fire: Prefix

addon/styles/_frost-numeric-select.scss
```diff
@@ -0,0 +1,99 @@
...
+    &:not(.disabled) {
+      outline: 0;
+      box-shadow: none;
+    }
+    .frost-icon-frost-numeric-chevron {
+      color: $arrow-focus-color;
+    }
+  }
+}
+.plusButton {
```
sglanzer-fire: There likely isn't a big different between `minus` and `plus` CSS, so I'd recommend writing it like

```
.minusButton,
.plusButton {
  // Common CSS
}

.minusButton {
  // Specific CSS
}

.plusButton {
  // Specific CSS
}
```

addon/styles/_frost-numeric-select.scss
```diff
@@ -0,0 +1,99 @@
...
+  font-size: 14px;
+  height: 14px;
+  width: 25px;
+  padding: 0px 6.2px 0px 6px;
```
sglanzer-fire: UX likes to stick to units of 5 when possible - also, even if they recommend `6px`, I wouldn't use decimal pixels (`6.2`) - it feels overly precise and not all browsers will respect it

addon/templates/components/frost-numeric-select.hbs
```diff
@@ -0,0 +1,77 @@
...
+      {{frost-button
+        hook='arrowDown'
+        class='minusButton'
+        icon='numeric-chevron'
+        pack='frost'
+        onClick= (action '_onMinusClick')
+      }}
+    {{/if}}
+  </div>
+  {{#if showError}}
```
sglanzer-fire: We _may_ need a generic error - e.g. "Hey, this field needs a value greater than 20" and the numeric error - I'd say the numeric error (`Input must be numeric`) would take precedence, so maybe we can do a computed property `_errorMessage` where the `errorMessage` (external) and the `_numericErrorMessage` (internal) are inputs to the computed property (CP) and we return the `_numericErrorMessage` if it's present.

I think that could remove the need for `showError` and the condition could just be `{{#if _errorMessage}}`

sglanzer-fire (changes requested):
Good first pass!  I'd also like to see some tests added, if you would like some help with testing @juwara0 is available to mentor

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+
+  // == Keyword Properties ====================================================
+  layout,
+
+  // == PropTypes =============================================================
+
+  propTypes: {    // options
+    align: PropTypes.string,
+    isHookEmbedded: PropTypes.bool,
+    receivedHook: PropTypes.string,
```
job13er: These two are obsolete now with the base frost component supporting `hookPrefix`, `frost-text` hasn't been updated to use `hookPrefix` yet because it would be a breaking chance, but new components should certainly not use these two.

job13er (dismissed):

addon/components/frost-numeric-select.js
sglanzer-fire: I wasn't entirely sure on that, probably shouldn't have used `frost-text` as my reference material for thinking about it

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+
+export default Component.extend({
+  // == Dependencies ==========================================================
+
+  // == Keyword Properties ====================================================
+  layout,
+
+  // == PropTypes =============================================================
+
+  propTypes: {    // options
```
dafortin: `// options` should be on the next line.

addon/components/frost-numeric-select.js
dafortin: I will expect this to be done internally

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+      tabindex: 0,
+      type: 'text'
+    }
+  },
+
+  // == Functions =============================================================
+
+  _isNumberKey (e) {
+    let decimalExist = this.get('value') && this.get('value').indexOf('.') >=0
+    let charCode = (e.which) ? e.which : e.keyCode
```
dafortin: We can do something like this `e.which || e.keyCode` if `e.which` is undefined/null you will get the value of `e.keyCode`

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+      spellcheck: false,
+      tabindex: 0,
+      type: 'text'
+    }
+  },
+
+  // == Functions =============================================================
+
+  _isNumberKey (e) {
+    let decimalExist = this.get('value') && this.get('value').indexOf('.') >=0
```
dafortin: I will prefer to have a `const value = this.get('value')` instead of calling get twice.

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+        return false
+      } else if (charCode != 46 && charCode > 31
+        && (charCode < 48 || charCode > 57))
+         return false
+
+      return true
+    }
+  },
+
+  //calculate how many digit a value has
```
dafortin: I will recommend using JsDoc format.

addon/components/frost-numeric-select.js
dafortin: Is this function failing the eslint complexity?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+  _calculateDigit (value) {
+    return value.slice(value.indexOf('.')+1).length
```
dafortin: `value.indexOf('.') + 1`

dafortin: Do we want to support only the `.` digit mark? https://en.wikipedia.org/wiki/Decimal_mark

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+  },
+
+  //set calculated and properly formated value
+  _setCalculatedValue (value, numberOfDigit) {
+    if(this.$('input').val().indexOf('.') > -1) {
```
dafortin: If I have the following:
1.00 => 1.00
1.000 => 1.00
1 => 1

Is that what is expected?

dafortin: We should probably add a function to determine if it's a number with decimals and do this check there instead of copying this check everywhere.

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+   */
+  @computed('isArrowVisible')
+  class (isArrowVisible) {
+    return isArrowVisible ? 'frost-numeric-select-button' : 'frost-numeric-select'
+  },
+
+  // == Lifecycle Hooks =======================================================
+  init () {
+    this._super(...arguments)
+    this.set('receivedHook', this.get('hook'))
```
dafortin: not necessary as @job13er mentioned.

addon/components/frost-numeric-select.js
dafortin: Can you create a variable to store the value?

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+        event.id = this.get('elementId')
+        event.value = event.target.value
+        this._eventProxy.input(event)
+      }
+    },
+    _onPlusClick (event) {
+      if (this.$('input').val()){
+        //calculate how many digit a value has
+        let numberOfDigit = this._calculateDigit(this.$('input').val())
+        var value = parseFloat(this.get('value'))+1
```
dafortin: Why are we adding `+ 1` to the value?

dafortin: ` + 1`

addon/components/frost-numeric-select.js
```diff
@@ -0,0 +1,183 @@
...
+          this._setCalculatedValue(value,numberOfDigit)
+        }
+      }
+    },
+    _onMinusClick (event) {
+      if (this.$('input').val()){
+        let numberOfDigit = this._calculateDigit(this.$('input').val())
+        let value = parseFloat(this.get('value'))-1
+
+        if (!this.get('isAllowUseDownButtonToNegative')) {
```
dafortin: It might be interesting to disable the minus button instead of doing something like this?
We might simply just not decrease if the number is  === 0

dafortin:
Great job :) Only a few small comments.

dafortin:
Opps closed the PR by mistake :(

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed. The component is named 'frost-numeric-field' now.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed. `isArrowVisible` prop has been already changed to `isIncrementControlVisible`.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: So when the user is allowed to input negative number, they can also use the down button to get to the negative number?  And when the user is not allowed to input negative number, he can't use the down button to get them?

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed. This condition is now removed.

addon/styles/_frost-numeric-select.scss
helloWorldCindy: Fixed.

addon/styles/_frost-numeric-select.scss
helloWorldCindy: Fixed.

addon/styles/_frost-numeric-select.scss
helloWorldCindy: Fixed.

addon/styles/_frost-numeric-select.scss
helloWorldCindy: Fixed.

addon/styles/_frost-numeric-select.scss
helloWorldCindy: Fixed.

addon/styles/_frost-numeric-select.scss
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed. It is removed now.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Added a computed property to disable button when the number is less or equal to zero. Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: This function is used for prevent floating point number precision in JS calculation. (e.g 1.22 - 1 will equals 0.199998, now it will be 1.22 - 1 = 0.22). I am not sure if it is required though.

addon/components/frost-numeric-select.js
helloWorldCindy: Can I use those key code in onKeyPress function? Seems like the key code in the event params is ASCII code.

addon/components/frost-numeric-select.js
job13er: Is 0 a valid key code? Probably not, so urn should be safe, just remember, if it's anything falsy it'll move to the other side of the ||, including `0`, `''`, or `false`. I know we've had at least one bug where a `false` value wasn't being handled properly because of a check like that. Just a reminder to be cautious.

addon/styles/_frost-numeric-select.scss
job13er: Also, aren't we using `kebab-case` for class names? So shouldn't it be `minus-button`, `plus-button`? The sass lint should actually complain about that (IIRC).

job13er: Looks like it does: 
```
  81:2  error  Class '.plusButton' should be written in lowercase with hyphens  class-name-format
```
Also, you have a lot of js lint and sass lint errors, which is why the build is failing. You can run: 
```
$ lint-all-the-things
```
at the root of the `ember-frost-core` repo to see the list of all the errors. Ideally, you'll want to configure your editor to show linting errors, then you can just open the files referenced in the results of `lint-all-the-things` and fix the issues your editor shows in each one. Particularly for property order linting of sass, not having it displayed in your editor will make it much more painful to fix.

sglanzer-fire: Yeah, we're using kebab-case

sglanzer-fire: Cindy didn't have sass-linting in her editor, it's on now

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/templates/components/frost-numeric-select.hbs
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-select.js
helloWorldCindy: Fixed.

addon/components/frost-numeric-field.js
```diff
@@ -0,0 +1,232 @@
...
+ */
+  _calculateDigit (value) {
+    return value.slice(value.indexOf('.') + 1).length
+  },
+
+  /** format decimal
+    * @param {string} value - value of the input
+    * @param {number} numberOfDigit - number of digit
+ */
+  _setCalculatedValue (value, numberOfDigit) {
```
sglanzer-fire, line 117: I'd probably suggest renaming this to reflect that it's not a `set` now, but we can clean that up the next time we're in the code

sglanzer-fire (approved):

sglanzer-fire:
Approved - the linting is failing your build currently, so you'll need to fix that, good otherwise

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/449/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

tests/dummy/app/pods/numeric-field/controller.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember'
+const {Controller} = Ember
+
+export default Controller.extend({
+  queryParams: ['value'],
+  value: '3',
```
sglanzer-fire: Sorry, I just noticed this - the value for a numeric field should really be in a numeric format - is the field capable of handling something other than a string (a number)?

sglanzer-fire (dismissed):
I just need confirmation that this can handle numeric values rather than string values

rox163:
I'm guessing this is still a component we want? @job13er @helloWorldCindy @dafortin  @juwara0 
Would like to see some movement on this PR accordingly.

job13er:
@rox163 I honestly don't know. I don't know the use-case it's trying to solve. I personally don't have any need of a "numeric selection component" although that description doesn't really say much about what it is and why it's needed/wanted. I'm guessing it's like `frost-select` but where the values can be numbers and not just strings? I've always just used `parseInt()` and a regular `frost-select` when working with a select that was choosing a number.

juwara0:
@rox163 @job13er - I found this past PR https://github.com/ciena-frost/ember-frost-core/pull/376 that was closed in favor of implementing a separate component. There is a little more information in the comments. I found this issue: https://github.com/ciena-frost/ember-frost-core/issues/405 that @sglanzer opened to track the task.

job13er:
@juwara0 Thanks for the context. I'd remove any reference to "select" in the PR description and changelog then, as it's completely misleading. The component name is `frost-numeric-field` which I'm not sure how I feel about, since it's basically `frost-text` but for a number, would `frost-number` be more appropriate? Naming aside, I have no problem with the component being created. I've just used `frost-text` for numeric inputs thus far.

rox163:
@helloWorldCindy could you please follow up with any remaining comments in the PR and also address the comments here https://github.com/ciena-frost/ember-frost-core/issues/405 and push updates please? Thanks

helloWorldCindy:
@rox163 Sure, will do it ASAP

addon/components/frost-numeric-field.js
```diff
@@ -184,6 +187,7 @@ export default Component.extend(FrostEventsProxyMixin, {
 
   // == Lifecycle Hooks =======================================================
   init () {
+    // this.set('value', String(this.get('value')))
```
notmessenger: Remove

helloWorldCindy: Fixed.

srowhani (approved):

NickLewanowicz (approved):
Great work Cindy, Amazon should be proud!

PR #452, opened 2017-05-19
Added expand and error components

AdamWard1995:

- [X] #minor# - adding functionality in a backwards-compatible manner

Main reason for this PR is to contribute the `frost-error` component so that there can be a common error page whenever an HTTP error is encountered. Made the `frost-expand` component for the show/hide details dropdown and made it a separate component for potential future use.

# CHANGELOG

* **Added** `frost-expand` component and `frost-error` component


tests/dummy/app/pods/error/template.hbs
```diff
@@ -0,0 +1,47 @@
...
+        <br/>
+        Controller
+        {{format-markdown "```javascript
+queryParams: {
+  showDetails: {
+    replace: true
+  }
+},
+showDetails: false,
+```"}}
```
job13er: Can we use `{{code-snippet'}}` instead of this? Then you know the snippet is working, if it comes from the live demo. It also lets the snippet be linted if it exists in a `.js` file.

job13er: Same with all the other instances of `format-markdown`

job13er (dismissed):

job13er:
Can we please get some additional information in the PR description about what `frost-error` and `frost-expand` are and why they are needed? For example, I thought our error messages are generally displayed in a `frost-notification` message. 

To quote @sglanzer 

>*please* be descriptive about the use case for the PR - the _why_ is just as important as the feature itself and not everyone reading the PR has the same background information that you do.  It’s difficult to make decisions that can potentially impact our application suites and the patterns of development without knowing why a feature was necessary.  It also helps immensely in the prioritizing of reviews for anyone who needs to read a large number of PRs on a daily basis.  Thanks in advance!

AdamWard1995:
What the frost-error component looks like (frost-expand is used for the 'Show details' dropdown, just made it it's own component):
screen shot 2017-05-19 at 8 25 38 am
screen shot 2017-05-19 at 8 25 26 am

AdamWard1995:
@job13er provided images in above comment and provided description in PR message

job13er:
I was unaware we're supposed to redirect to a new page when an API error is encountered. Is that only on 500 errors, I'm assuming. I'd hate to be redirected to an error page because of a duplicate name or something. Also, I believe we already had a collapsible panel component, I guess we should deprecate that guy in favor of this one? When is an app developer supposed to redirect to a whole error page vs showing the notification message?

sglanzer-fire:
@job13er the error page is for transition errors (failure of a model hook) - so the base level entity being requested fails.  Notifications are used for anything else - e.g. failed to retrieve secondary data, failed to perform an in-page operation, etc.

The collapsible panel was our fork of `ted` code, right?  We don't have to deprecate immediately, but @AdamWard1995 can you raise an issue to evaluate both and make sure our functionality covers the use cases?  We can deprecated after that.

AdamWard1995:
@sglanzer I had done a quick look to see if there was a similar UI component and didn't find one so no, I coded it from scratch

job13er:
Here it is: https://github.com/ciena-frost/ember-frost-collapsible-panel

Not sure why that wouldn't come up in a search. I suppose if you searched for "expand" it wouldn't, but a casual browse through our existing repos (we don't have *that* many) should have revealed it.

sglanzer-fire:
Yeah, that's the one that's a fork of `ted`'s collapsible panel

AdamWard1995:
So obviously cannot use that existing component in this error component as it would create a dependency cycle. @sglanzer want me to get rid of the separate expand component I introduced and just move the needed code to the error component?

sglanzer-fire:
@AdamWard1995 no, I'm fine with taking the expand component back in-house instead of using TED's, but let's raise an issue to look at the deprecation and not move actively on that yet

AdamWard1995:
Ok, raise the issue in this repo?

sglanzer-fire:
Yeah, that'd be good

addon/components/frost-error.js
```diff
@@ -0,0 +1,62 @@
...
+import Component from './frost-component'
+import {PropTypes} from 'ember-prop-types'
+
+export default Component.extend({
+  // == Dependencies ==========================================================
+
+  // == Keyword Properties ====================================================
+
+  layout,
+  tagName: 'div',
```
sglanzer-fire: Just a "learning" note - components have `div` as the tagName by default

addon/components/frost-error.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * Component definition for the frost-error component
```
sglanzer-fire: I think it would make sense to have the name of this a bit more specific to http error codes.  The way it's structured right now is rather specific and it probably makes more sense to have error presentations for specific cases rather than trying to do one generic "handle everything" version

How about something like `frost-ajax-error-page`?

addon/components/frost-error.js
```diff
@@ -0,0 +1,62 @@
...
+   * passed in/overwritten.
+   */
+  propTypes: {
+    // options
+    description: PropTypes.string,
+    suggestion: PropTypes.string,
+    errorCode: PropTypes.number,
+    errorTitle: PropTypes.string,
+    errorMessage: PropTypes.string,
+    errorDetails: PropTypes.string,
```
sglanzer-fire: I think most of the above should be required, right?  Maybe we could get away with `suggestion` (others?) being optional, but it probably doesn't make sense to not require information that would always be available in the case of an HTTP response

addon/components/frost-error.js
```diff
@@ -0,0 +1,62 @@
...
+    showDetails: PropTypes.bool
```
sglanzer-fire: In this particular case I don't see a lot of utility in holding the expand/collapse state outside of component (retain the expansion when the user refreshes an error page?) - so I would suggest making `showDetails` internal to reduce the footprint of the interface for a consumer.

addon/components/frost-error.js
```diff
@@ -0,0 +1,62 @@
...
+  /** @returns {Object} the default property values when not provided by consumer */
+  getDefaultProps () {
+    return {
+      // options
+      description: '',
+      suggestion: '',
+      errorCode: -1,
+      errorTitle: '',
+      errorMessage: '',
+      errorDetails: '',
```
sglanzer-fire: As noted above, I think most of these should be required, so we can drop the defaults - something like a `-1` kind of confirms that assumption I believe.

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+import {PropTypes} from 'ember-prop-types'
+
+export default Component.extend({
+  // == Dependencies ==========================================================
+
+  // == Keyword Properties ====================================================
+
+  layout,
+  tagName: 'div',
+  classNameBindings: ['cssStateClass'],
```
sglanzer-fire: There is a neat syntax for classNameBindings that actually lets you specify true/false values `condition:trueValue:falseValue`, so I'd suggest simplifying this to:

`classNameBindings: ['expanded:expanded:collapsed']`

And you could drop the associated computed property

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Properties for this component. Options are expected to be (potentially)
+   * passed in to the component. State properties are *not* expected to be
+   * passed in/overwritten.
+   */
+  propTypes: {
+    // options
+    onExpand: PropTypes.func,
+    onCollapse: PropTypes.func,
```
sglanzer-fire: Can we collapse these events into one `onChange` event?  Effectively we're just going to be binding a toggled boolean back to a local property, so distinguishing between the two cases shouldn't be necessary

sglanzer-fire: Also, I believe `onChange` should be required

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+    expanded: PropTypes.bool,
```
sglanzer-fire: Required

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+    duration: PropTypes.string,
```
sglanzer-fire: Is duration a string?  I would have expected a number (milliseconds) - also, maybe we want to make the naming a bit more explicit - `animationDuration`?

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+    // state
+  },
+
+  /** @returns {Object} the default property values when not provided by consumer */
+  getDefaultProps () {
+    return {
+      // options
+      onExpand () {},
+      onCollapse () {},
+      expanded: false
```
sglanzer-fire: Defaults for the `expandLabel` (Expand) and `collapseLabel` (Collapse)

sglanzer-fire: Also, we don't want default for external bindings (goes with the `isRequired` comment above)

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+   */
+  propTypes: {
+    // options
+    onExpand: PropTypes.func,
+    onCollapse: PropTypes.func,
+    expanded: PropTypes.bool,
+    duration: PropTypes.string,
+    expandLabel: PropTypes.string,
+    collapseLabel: PropTypes.string,
+    content: PropTypes.oneOfType([
```
sglanzer-fire: I think you're ok with `any` here - you've already got the check for a `component` later and anything else should just render out as-is (number, string, etc.)

sglanzer-fire: Also, thanks for the including the `block` format too!

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+
+  @readOnly
+  @computed('content')
+  /**
+   * Determine whether or not the content is a component
+   * @param {*} content - content to display
+   * @returns {Boolean} whether or not the content is a component
+   */
+  isComponentContent (content) {
+    return content !== null && typeof content !== 'string'
```
sglanzer-fire: We probably want to use the same type of validation that `prop-types` is using https://github.com/ciena-blueplanet/ember-prop-types/blob/master/addon/utils/validators/ember-component.js

In fact, if we could use that validator directly it would be safer

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+
+  @readOnly
+  @computed('duration')
+  /**
+   * Determine duration of expansion animation
+   * @param {String} duration - duration string
+   * @returns {Number} duration in milliseconds of animation
+   */
+  durationTime (duration) {
+    if (duration === 'fast') {
```
sglanzer-fire: I'd suggest skipping the enum here and just letting the consumer specify a value directly

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+      return 250
+    } else if (duration === 'slow') {
+      return 1000
+    } else if (duration === 'none') {
+      return 0
+    }
+    return 500
```
sglanzer-fire: Generally I would default to `300` for most animations (the edge of perceived "instant") https://medium.com/@jakob_anderson/speed-performance-and-human-perception-70ae83ea144e

addon/components/frost-expand.js
sglanzer-fire: Default for `duration`

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+  },
+
+  // == Functions =============================================================
+
+  updateVisibility (expand, duration = 0) {
```
sglanzer-fire: Would you mind adding a `// TODO` here - a really nice future enhancement would be to measure the content size and adjust the animation duration accordingly (if no explicit duration is provided) - this would also be the point in time where re-introducing your enum concept (fast, medium, slow) would make sense, but I'd probably call it `animationSpeed` or something like that (duration implying an explicit control)

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+    if (expand) {
+      this.$().find('.frost-expand-scroll').slideDown(duration)
+    } else {
+      this.$().find('.frost-expand-scroll').slideUp(duration)
+    }
+  },
+
+  stateChanged: observer('expanded', function () {
```
sglanzer-fire: https://www.youtube.com/watch?v=7PUX27RKCq0 is worth a watch on the topic of `observers`

In this case I would recommend avoiding the observer by using the `didReceiveAttrs` hook.  Effectively you'll keep a local copy of `_expanded` and if `expanded` changes you would set the value of `_expanded` and trigger the animation

```
didReceiveAttrs() {
  const _expanded = this.get('_expanded')
  const expanded = this.get('expanded')

  if (_expanded !== expanded) {
    this.set('_expanded', expanded)
    this.updateVisibility...
  }
}
```

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+  // == DOM Events ============================================================
+
+  // == Lifecycle Hooks =======================================================
+
+  didInsertElement (model, transition) {
+    const startExpanded = this.get('expanded')
+    if (!startExpanded) {
+      this.updateVisibility(startExpanded)
+    }
+  },
```
sglanzer-fire: The `didReceiveAttrs` comment above should be able to cover this as well

addon/components/frost-expand.js
```diff
@@ -0,0 +1,139 @@
...
+        if (this.onExpand) {
+          this.onExpand()
+        }
+      } else {
+        if (this.onCollapse) {
+          this.onCollapse()
+        }
+      }
+
+      this.set('expanded', expand)
```
sglanzer-fire: Don't set state internally, rely on the DDAU (data down, actions up) pattern - with the suggestion to use `onChange` this would be simplified to `this.onChange(!this.get('expanded'))`

addon/styles/_frost-expand.scss
```diff
@@ -0,0 +1,56 @@
+$frost-color-expand-label-text: $frost-color-blue-1 !global;
+$frost-color-expand-scroll-border: $frost-color-lgrey-3 !global;
+$frost-color-expand-content-bg: $frost-color-lgrey-4 !global;
+$frost-color-expand-content-text: $frost-color-grey-3 !global;
+
+.frost-expand {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
sglanzer-fire: Not picturing this, can you give a brief description of why space-between is the right choice?

addon/styles/_frost-expand.scss
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  &.expanded {
+    .frost-expand-label-icon {
+      transform: rotate(0deg);
+    }
+  }
+
+  &-label {
```
sglanzer-fire: Relatively new (personal) convention, but I'd prefer this to be `frost-expand-label` (easier to find the rule)

Same would apply for text, icon, scroll and content below

addon/styles/_frost-expand.scss
```diff
@@ -0,0 +1,56 @@
...
+    display: flex;
+    flex-direction: row;
+    flex-shrink: 0;
+    justify-content: flex-start;
```
sglanzer-fire: I assume the main case for flex here is to keep the text and icon centered?  Maybe you can provide some additional info, but I'd expect to see an `align-items: center` here

addon/styles/_frost-expand.scss
```diff
@@ -0,0 +1,56 @@
...
+
+    &-icon {
+      width: 20px;
+      height: 20px;
+      transition: transform .2s;
+    }
+  }
+
+  &-scroll {
+    border: 1px ridge $frost-color-expand-scroll-border;
```
sglanzer-fire: Can you validate `ridge` with the visual designers?  I don't think it's going to get sign-off, but I could be wrong

addon/styles/_frost-expand.scss
```diff
@@ -0,0 +1,56 @@
...
+
+    &-content {
+      height: 100%;
```
sglanzer-fire: Is there a `position: relative` parent (I may have missed it)?

addon/templates/components/frost-error.hbs
```diff
@@ -0,0 +1,23 @@
+<div class="frost-error-description" data-test={{hook (concat hook '-description')}}>
```
sglanzer-fire: When we build up a child hook using `concat` I'd like to concat the `hookPrefix` instead of `hook`

sglanzer-fire: `Component`s extending from the `ember-frost-core` component have a nice property named `css` that automagically get the name of the component so you can write the class as:

`<div class='{{css}}-description'...`

Which protects you if the component is renamed in the future (plus it's shorter)

If you do the same for a child component the format is
```
{{some-component
  class=(concat css '-something')
```

addon/templates/components/frost-error.hbs
```diff
@@ -0,0 +1,23 @@
...
+<div class="frost-error-http-error" data-test={{hook (concat hook '-http-error')}}>
+  Error {{errorCode}} - {{errorTitle}}
+</div>
+<div class="frost-error-message" data-test={{hook (concat hook '-message')}}>
+  {{errorMessage}}
+</div>
+{{#frost-expand
+  class="frost-error-expand"
+  hook=(concat hook '-expand')
+  expanded=false
```
sglanzer-fire: This should be bound to the state property for expansion in the `frost-error` component js

addon/templates/components/frost-error.hbs
```diff
@@ -0,0 +1,23 @@
...
+  duration="fast"
+  expandLabel="Show details"
+  collapseLabel="Hide details"
+  expanded=showDetails
```
sglanzer-fire: Duplicate declaration of `expanded`

sglanzer-fire: Missing `onFoo` action binding?

addon/templates/components/frost-expand.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="frost-expand-label"  data-test={{hook (concat hook '-label')}} onClick={{action '_onClick'}}>
+  {{frost-icon
+    class='frost-expand-label-icon'
```
sglanzer-fire: `frost-icons` have an automagical class created based on the pack and icon

`frost-icon-<pack>-<icon>` so this instance should already have a `frost-icon-frost-expand-collapse` class available

Also, any reason we aren't re-using `chevron` here?

addon/templates/components/frost-expand.hbs
```diff
@@ -0,0 +1,29 @@
...
+    hook=(concat hook '-label-icon')
+    pack='frost'
+    icon='expand-collapse'
+  }}
+  <div class="frost-expand-label-text" data-test={{hook (concat hook '-label-text')}}>
+    {{#if expanded}}
+      {{if collapseLabel collapseLabel 'Collapse'}}
```
sglanzer-fire: I made a comment above, but the default of `'Collapse'` makes the most sense to put in the default values in the prop-types area of the code so that all the information is centralized

tests/dummy/app/pods/error/controller.js
```diff
@@ -0,0 +1,16 @@
...
+const {Controller} = Ember
+
+export default Controller.extend({
+  // BEGIN-SNIPPET error-query-params
+  queryParams: {
+    showDetails: {
+      replace: true
+    }
+  },
+  showDetails: false,
```
sglanzer-fire: Noted above, but I don't believe it's necessary to maintain this state externally - I'm open to someone telling me I'm wrong though :)

tests/dummy/app/pods/expand/template.hbs
```diff
@@ -0,0 +1,245 @@
...
+        {{! END-SNIPPET expand-events}}
+      </div>
+      <br/>
+      <div class="snippet">
+        {{code-snippet name="expand-events.hbs"}}
+      </div>
+    </div>
+  </div>
+
+  <div class='section-title'>Query Parameter</div>
```
sglanzer-fire: Don't worry about providing a specific QP example - this is an option available to any case where external state is held and consumers should be familiar with this as an Ember option

tests/integration/components/frost-error-test.js
```diff
@@ -0,0 +1,114 @@
...
+    beforeEach(function () {
+      this.render(hbs`
+        {{frost-error
+          hook="fe"
+        }}
+      `)
+    })
+
+    it('should add hook to description', function () {
+      expect($hook('fe-description')).to.have.class('frost-error-description')
```
sglanzer-fire: I'd suggest using variables for the classes, find/replace isn't terrible, but a prefix variable is better

sglanzer-fire (changes requested):
As per comments - ping me if you want to discuss anything / when the next version is ready for a review pass

addon/components/frost-expand.js
AdamWard1995: I agree with collapsing it into one event but should it be required? It isn't for the toggle button

addon/components/frost-expand.js
AdamWard1995: Thanks, saw you guys didn't use this anywhere else so assumed there had to be a better way

addon/components/frost-expand.js
sglanzer-fire: Toggle is probably wrong in that case - there isn't much utility to an input component that can't handle input

AdamWard1995: Ok, but this isn't an input component. For example, when using it in the error page, do we really care about each time it is expanded or collapsed? We're just providing some content to the user that they can choose whether they want to view or not but we don't care about every time they toggle the visibility of the content. Just my 2 cents, I'm fine with doing it either way

addon/styles/_frost-expand.scss
AdamWard1995: Purpose was to avoid this: 
screen shot 2017-05-26 at 11 30 50 am

is there a better way?

addon/templates/components/frost-expand.hbs
AdamWard1995: Don't see how the `chevron` is different from the `expand-collapse` one other than the `expand-collapse` one is bigger

tests/dummy/app/pods/error/controller.js
AdamWard1995: Shouldn't be external but can make this boolean an internal state variable based on our conversation this morning right?

addon/styles/_frost-expand.scss
AdamWard1995: Forgot to remove when I was playing around with the flexbox stuff

addon/styles/_frost-expand.scss
AdamWard1995: Replaced with `display: inline-flex;` unless there is a better way

sglanzer-fire: I'm fine with flex, but I think it would make sense to add `align-items: center` to make the behaviour explicit (and guard against a different height text/icon)

AdamWard1995: Ya ok, sorry keep getting `align-items` and `justify-content` confused

addon/templates/components/frost-expand.hbs
sglanzer-fire: It's an SVG so resize is par for the course - we should probably just use chevron here so that we don't end up with a bunch of dupes of the same icon

addon/styles/_frost-ajax-error-page.scss
```diff
@@ -0,0 +1,42 @@
...
+
+  &-message {
+    margin-bottom: 20px;
+    color: $frost-color-error-server-message;
+    font-size: $frost-font-s;
+  }
+
+  &-expand {
+    width: 600px;
+    max-height: 400px;
```
sglanzer-fire, line 36: We could probably make this more dynamic in the future, but it's fine for now

sglanzer-fire (approved):

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/452/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #466, opened 2017-06-16
Add style rules for hover state of frost-toggle switch

jscharf:

# CHANGELOG

Added style rules for the hover state of the frost-toggle button switch. The switch will now turn a light blue color when the user mouses over the toggle. If the toggle is disabled, the switch will not change color.

### DIFF
--- a/addon/styles/_frost-toggle.scss
+++ b/addon/styles/_frost-toggle.scss
@@ -5,6 +5,9 @@ $frost-toggle-button-switch-height: 27px;
 $frost-toggle-off-color: #000;
 $frost-toggle-on-color: #fff;
 
+$frost-toggle-button-hover-color: #c6eafc;
+$frost-toggle-button-hover-shadow: rgba(0, 0, 0, .2);
+
 .frost-toggle {
   display: inline-flex;
   align-items: center;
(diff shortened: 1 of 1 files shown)


sglanzer-fire:
@jscharf thanks :)

Would you mind posting screenshots (can past directly into a comment) and @baconC can review?

jscharf:
Sure! This is what it looks like now if the toggle is off and the user hovers over the switch:
frost-toggle before
And this is what it looks like after my changes:
frost-toggle after
Similarly, this is what it looks like now if the toggle is on and the user hovers over the switch:
frost-toggle-on before
And this is what it looks like after my changes:
frost-toggle-on after

addon/styles/_frost-toggle.scss
```diff
@@ -58,9 +54,8 @@ $frost-toggle-button-hover-shadow: rgba(0, 0, 0, .2);
     &:after {
       display: block;
       position: absolute;
-      width: 40px;
+      width: 45%;
```
sglanzer-fire, line 57: Is this width constrained by a relative parent?

jscharf:
Here is what the toggle looks like now with latest CSS changes.
Off state:
frost-toggle
On state:
frost-toggle-on

baconC:
APPROVED

addon/styles/_frost-toggle.scss
sglanzer-fire: Answered my own question by scrolling up ;)

sglanzer-fire:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/466/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

addon/styles/_frost-toggle.scss
jscharf: Yes it's constrained by the parent element which in this case is a label with the `.frost-toggle-button` class. If the width was 100% it would span the entire width of the toggle.

PR #477, opened 2017-07-20
Change from Jquery create canvas element to document.create

EWhite613:
JQuery create behaves strangley in Firefox sometimes where it seems like the canvas doesn't have a surface

# CHANGELOG

* The JQuery creation of a canvas element was behaving oddly in some apps on Firefox.

### DIFF
--- a/addon/utils/text.js
+++ b/addon/utils/text.js
@@ -1,7 +1,7 @@
 import Ember from 'ember'
 const {$} = Ember
 
-const ctx = $('<canvas />').get(0).getContext('2d')
+const ctx = document.createElement('canvas').getContext('2d')
 
 /**
  * Get width of text for a specific font


quincyle:
Approved

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/477/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

EWhite613:
Using the JQuery create element on firefox would produce errors in some apps that resulted in simple tasks like measuring text resulting in a canvas error.
```
$('<canvas />').get(0).getContext('2d').measureText('foo')
[Exception... "Failure"  nsresult: "0x80004005 (NS_ERROR_FAILURE)"  location: "JS frame :: debugger eval code :: <TOP_LEVEL> :: line 1"  data: no]
```

As well setting the font here https://github.com/ciena-frost/ember-frost-core/blob/master/addon/utils/text.js#L83 results in the same issue

Only after the jquery canvas is appended to the body will it fix the error (if the canvas element is removed immediatly after the issue still remains solved).

 It seems as if JQuery somehow has the canvas uninitialized. I believe the error comes from here https://dxr.mozilla.org/mozilla-central/source/dom/canvas/CanvasRenderingContext2D.cpp#4524

As the Firefox's JQuery Canvas Element's `ownerDocument` is  belongs to `about:blank` with a very plain document. 
Whereas in Chrome the element belongs to the Ember application document (same goes for using `document.createElement` in Firefox)

EWhite613:
The JQuery's ownerDocument is set with this [logic](http://api.jquery.com/jquery/#jQuery-html-ownerDocument)

> By default, elements are created with an .ownerDocument matching the document into which the jQuery library was loaded. Elements being injected into a different document should be created using that document, e.g., $("<p>hello iframe</p>", $("#myiframe").prop("contentWindow").document)

PR #511, opened 2017-10-31
Make run loop aware adjustments

EWhite613:

# CHANGELOG

* Fix frost-select-dropdown breaking in tests because `autorun loop is disabled`
* Add pre-cautionary run-loop on frost-scroll onMouseEnter

### DIFF
--- a/addon/components/frost-scroll.js
+++ b/addon/components/frost-scroll.js
@@ -70,9 +70,7 @@ export default Component.extend({
     }
 
     this._mouseEnterHandler = () => {
-      run(() => {
-        this.get('onMouseEnter')(this.get('element'))
-      })
+      this.get('onMouseEnter')(this.get('element'))
     }
 
     if (typeOf(this.onScrollUp) === 'function') {
--- a/addon/components/frost-scroll.js
+++ b/addon/components/frost-scroll.js
@@ -70,7 +70,9 @@ export default Component.extend({
     }
 
     this._mouseEnterHandler = () => {
-      this.get('onMouseEnter')(this.get('element'))
+      run(() => {
+        this.get('onMouseEnter')(this.get('element'))
+      })
     }
 
     if (typeOf(this.onScrollUp) === 'function') {
--- a/addon/components/frost-select-dropdown.js
+++ b/addon/components/frost-select-dropdown.js
@@ -2,7 +2,6 @@
  * Component definition for frost-select-dropdown component
  */
 import Ember from 'ember'
-const {$, deprecate, get, isArray, isEmpty, merge} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import {task, timeout} from 'ember-concurrency'
 import {PropTypes} from 'ember-prop-types'
@@ -13,6 +12,7 @@ import {keyCodes} from '../utils'
 import {trimLongDataInElement} from '../utils/text'
 import Component from './frost-component'
 
+const {$, deprecate, get, isArray, isEmpty, merge, run} = Ember
 const {DOWN_ARROW, ENTER, ESCAPE, TAB, UP_ARROW} = keyCodes
 
 const BORDER_HEIGHT = 1
@@ -254,14 +254,18 @@ export default Component.extend({
     Array.from(listItemElements).forEach((li, index) => {
       $(li)
         .mousedown(() => {
-          if (this.isDestroyed || this.isDestroying) return
-          const value = this.get(`items.${index}.value`)
-          this.set('focusedIndex', index)
-          this.send('selectItem', value)
+          run(() => {
+            if (this.isDestroyed || this.isDestroying) return
+            const value = this.get(`items.${index}.value`)
+            this.set('focusedIndex', index)
+            this.send('selectItem', value)
+          })
         })
         .mouseenter(() => {
-          if (this.isDestroyed || this.isDestroying) return
-          this.set('focusedIndex', index)
+          run(() => {
+            if (this.isDestroyed || this.isDestroying) return
+            this.set('focusedIndex', index)
+          })
         })
     })
   },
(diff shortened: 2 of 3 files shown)


addon/components/frost-scroll.js
```diff
@@ -70,7 +70,9 @@ export default Component.extend({
     }
 
     this._mouseEnterHandler = () => {
-      this.get('onMouseEnter')(this.get('element'))
+      run(() => {
```
notmessenger: This does not feel like the right way to solve anything.  The setting of this handler via `this.$().on('mouseenter', this._mouseEnterHandler)` is already scheduled via `run.scheduleOnce('afterRender',` and when the `mouseenter` event is being fired there is no run loop related activity occurring (meaning a mouse over occurred - there's not rerendering etc going on).   Have you exhausted the usage of this component in the application for which the tests are failing as well as the test setup?

notmessenger: I haven't checked the _addon/components/frost-select-dropdown.js_ file changes yet but suspect my comments will be the same

sophypal: Isn't it possible that the consumer of this addon would do `set` and `setProperties` inside of their `onMouseEnter` callback. If it wasn't inside a run loop, every `set` would trigger a render if it altered properties that affected computed properties it needed to render.

EWhite613: I took this one out, I don't see any usage's in our app so it doesn't really matter for us. 
The frost-select `mousedown`, `mouseenter`, and `_keyDownHandler` are 100% a problem tho. If I turn off the auto run loop I can reproduce the testing error consistently we intermittently see in our app from test cases that interact with `frost-select-dropown`

notmessenger:
@EWhite613 Did you remove the _addon/components/frost-scroll.js_ changes based on my comment or @sophypal's?  How were you able to so quickly determine that code you previously said was needed  to fix

> breaking in tests because autorun loop is disabled

is no longer needed?  You tested it that quickly in a consuming app?  And if so, and it was not needed, what led you to believe this was originally needed then?

I'm concerned that this PR, and subsequent changes, may not be being given the thoroughness it needs.

addon/components/frost-scroll.js
EWhite613: @sophypal that was my concern as well

notmessenger: @sophypal That is certainly a possibility though that onus should be on the application, as they are choosing to do things in their callback that is causing this issue.  If they chose not to do this at all, or did not update properties that would cause problems this problem would not exist.  And the properties they are choosing or not choosing to update are properties in their application code (or consumption therein) not any properties provided by this component, which again, makes it an application concern.

EWhite613:
@notmessenger I removed it based on your comment and the fact I never truly verified it. Our consuming app doesn't make use of that callback, it just looked suspect ( I was thinking the same thing @sophypal was "consumers could be doing things in the callback that required a run loop")

> breaking in tests because autorun loop is disabled

Was mainly directed at `frost-select-dropdown` because I could verify in our consuming app by disabling the autorun loop and running the same logic from our test cases (that cause `Error: Assertion Failed: You have turned on testing mode, which disabled the runloop's autorun. You will need to wrap any code with asynchronous side-effects in a run` ) that the async logic wasn't running in the Ember run loop based on the stack trace from the error.

addon/components/frost-scroll.js
sophypal: If the application was consuming a 3rd-party non-ember library, then yes the onus should be on them. However, they're importing an Ember addon which needs to be safe to use. Why would you expect all consumers of an "Ember addon" to assume that any callbacks you pass won't be async safe? Shouldn't the responsibility be on the addon for managing this compliance when it was the addon that imported the 3rd-party lib. How are consumers to know, without peeking through the code, that it was using native events?

sophypal: Using your logic @notmessenger, you're saying that any callback I pass to an Ember addon should be surrounded with `run`, if I need to set my local properties. That seems wrong.

notmessenger: It's not this add-on causing the problem though.  The problem is that the consuming application is attempting to modifying properties multiple times in the same render.  As we've discussed before the way to solve this is to adjust the code flow triggering computed properties, not to wrap everything in a run loop.  The point is this can be solved in the consuming application - not by using a run loop but by orchestrating a proper code flow.  This makes this issue not one of where the run loop logic should go - and that it should be in the add-on for all to use and not have think about implementing - but one of how to consume this functionality correctly.  The fact that we're changing this add-on to run changes to the consuming apps code at a different time to avoid changing its value in the same render means that its the app that's not orchestrated correctly.

notmessenger: And even if I were to agree that this should be done in the add-on, we're saying that `run()` is good enough scheduling?  What about all of the available run queues?  How are those to be coordinated with an application's needs?  They wouldn't be - it should be up to the app to handle the use of `run()`, and more correctly, to not need to use it at all.

sophypal: I understand your concern but it's actually pretty common that applications set properties during an action caused by some event.  The issue you're describing is when applications modify properties multiple times during a single render.  I don't see how a mouse event being handled would cause a double render if the event happens outside the render process.  Anyways, we maybe saying the same thing but not understanding each other.  I just find that generally, it's a good practice to guard event handling code that is not run loop aware in a run loop.

notmessenger:
@EWhite613 It does not appear that you have opened a PR in the application consuming this branch's code changes to ensure that this will actually fix things, since the errors I know you're experiencing only occur in the CI environment and not local development.  Or did you do so and delete the PR?

notmessenger:
@EWhite613 There are no test changes in this PR.

EWhite613:
@notmessenger the CI is so intermittent in the failures it's easier to reproduce locally with the autorun loop disabled in my dev environment where I can consistently reproduce. Then to confirm my fixes solved the issue all I had to do was an `npm link` and follow the same steps I used to consistently reproduce.

If I made a PR with the changes I wouldn't be able to know if I actually solved the issue or it was just one of the times it didn't intermittently fail


I'll add some tests to the logic that wasn't tested already

addon/components/frost-scroll.js
EWhite613: @notmessenger side note:
`Ember.run(...)` doesn't avoid the same render error. `Ember.run(..)`executes the callback immediately. You're thinking of `Ember.run.next(...)` that'll wait for the next render.
https://ember-twiddle.com/b9b2d1a1bb13a0ad86e202d0db3e57f1?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs

sophypal: That's an excellent point and a point I was trying to make as well. The use of a `run` loop inside an non-ember async event is to avoid unexpected async behavior not necessarily to avoid the render issue were we seeing scarcely in mcp. I totally agree with @notmessenger in that using `run.next` or `run.schedule` as a means to avoid the render error is not recommended but I didn't think @EWhite613 added it to avoid the render issues.

sophypal:
👍

addon/components/frost-scroll.js
notmessenger: Except you previously said @sophypal 

> If it wasn't inside a run loop, every set would trigger a render if it altered properties that affected computed properties it needed to render.

and @EWhite613 agreed, meaning that @EWhite613 added it to avoid the render issues.

EWhite613: @notmessenger I added it to avoid users from failing when the autorun loop is disabled. I used `run` not `run.next`, render issues would still be prevalent

sophypal: LOL. I did say that didn't I. I can see the misunderstanding now. When I said that, I didn't mean the specific multiple render error we've been seeing since the upgrade to Ember 2.12. I've come across issues, related to rendering, that are triggered due to 2 separate async events executing at similar timeframes. I incorrectly implied that using a run block would resolve the two events into one render, whilst not using a run block would trigger multiple renders--which is not the same as the double render error. Using `run.schedule`, will allow parallel stream of events to coalesce into one render call if they happen to execute before the next scheduled queue runs.

There are other unpredictable behaviors, but the issues I've seen in the past are observed as rendering issues (i.e. doing `set` outside a run loop), and they most often break our test because the auto run loop is disabled.

sophypal: To clarify even more, the double render error I mentioned is errors due to setting properties during a render event which is more like a cascade rendering issue. The other rendering issue I've been discussing occur as 2 separate rendering events due to 2 separate async events. I can totally see now the confusion I've caused.

PR #516, opened 2017-11-08
Added falseValue and trueValue to frost-checkbox options

TheOtherDude:

# CHANGELOG

* Added optional falseValue and trueValue properties to `frost-checkbox`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ coverage.json
 /typings/*
 .vscode
 **.ember-try
+yarn.lock
--- a/addon/components/frost-checkbox.js
+++ b/addon/components/frost-checkbox.js
@@ -22,9 +22,11 @@ export default Component.extend({
     autofocus: PropTypes.bool,
     checked: PropTypes.bool,
     disabled: PropTypes.bool,
+    falseValue: PropTypes.any,
     inputId: PropTypes.string,
     label: PropTypes.string,
-    size: PropTypes.string
+    size: PropTypes.string,
+    trueValue: PropTypes.any
 
     // state
   },
@@ -35,9 +37,11 @@ export default Component.extend({
       autofocus: false,
       checked: false,
       disabled: false,
+      falseValue: false,
       inputId: null,
       label: '',
-      size: 'small'
+      size: 'small',
+      trueValue: true
     }
   },
 
@@ -116,10 +120,11 @@ export default Component.extend({
      */
     input () {
       let id = this.get('value')
+      const {falseValue, trueValue} = this.getProperties(['falseValue', 'trueValue'])
       if (this.onInput) {
         this.onInput({
           id: isEmpty(id) ? this.get('elementId') : id,
-          value: this.$('input').prop('checked')
+          value: this.$('input').prop('checked') ? trueValue : falseValue
         })
       }
     }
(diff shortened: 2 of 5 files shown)


gknoy:
👍

cstolli:
LGTM, @notmessenger ?

notmessenger (dismissed):
I need some time to review this implementation.  I know for sure it is an undesired implementation - I need to time to write everything up.

markcole99:
@notmessenger - can you give us a timeline on getting this reviewed?

notmessenger:
@markcole99 Hopefully today

markcole99 (approved):

job13er:
@notmessenger What's so undesirable about this implementation? It's not exactly a huge PR. Taking the time to say "No" without saying anything about why "No" seems completely unhelpful.

notmessenger:
@job13er @gknoy Taking the time to say "No" is necessary because by the time I was even able to type that response in order to block the PR from being merged 3 individuals had already indicated their acceptance of this PR (2 on here on 1 in private Slack).  Had I not done that this code would have been merged before I could review it and provide the feedback that is necessary.  I will indicate what is so undesirable about this implementation when I have the time to properly dedicate to its review.

gknoy:
@notmessenger Perhaps that's an indicator that the things that are wrong with it are (a) not noticeable to several other engineers, or (b) minor enough that a follow-up patch to adjust implementation would be sufficient.  We need the capability, and this adds functionality that wasn't there before.

If you would rather see it done a different way, as long as the capability is still there then I propose that we move forward with this, and then you can either PR your suggested changes, or we can patch it with a follow-up PR.

notmessenger:
@cstolli @gknoy @job13er @markcole99 The reason for opening a PR is to receive a peer review of your code.  This review may take a little bit of time to perform, especially when some of the peers have noticed somethign that should be changed or is reviewing it from a different perpspective than others.  This amount of time does not seem to be any that I have directed this message too seems willing to wait for.  Instead the approval of PRs is a race to rubber stamp the changes.  We seemingly view the PR process a step that we must check off that gets in the way of us "rammin' and jammin'" instead of what it should be which is to receive feedback in order to improve the quality of our code.  If we took the time to discuss things before implementating them and getting "bogged down" in the PR process by individuals who voice concerns about the quality then this process would not be a blocker.   This is actually what our Senior Director directed us to do yesterday as well, incidentally.  And it's something I have been advocating for for the past year in my employment with this organziation and for years prior in my career.

I have now spent more time outlining how we should be doing things rather than being able to address the actual PR itself so I am I going to remove my objection.  Until we decide as an organiation to participate in the PR process I am not able to effectively influence code quality.

notmessenger:
@gknoy 

> Perhaps that's an indicator that the things that are wrong with it are (a) not noticeable to several other engineers

Which is the purpose of a PR.  You say this statement to insinuate that since I'm the only one having an issue with it it must be my issue that can brushed aside and not an actual concern that needs to be addressed.   Perhaps is that the other engineers do not see the same problem due to experience or their rush to rubber stamp PRs.

TheOtherDude:
Going to merge this now since everyone seems to be OK enough with it. If an issue arises with this implementation, I'll gladly do a follow up PR.

PR #530, opened 2017-12-20
Fixed dropdown not updating when select items change

TheOtherDude:

# CHANGELOG

* **Fixed** dropdown not updating with new select items


addon/components/frost-select.js
```diff
@@ -257,6 +257,19 @@ export default Component.extend({
...
+
+  @readOnly
+  @computed('items', 'selectedItems')
+  /**
+   * As of Ember 2.10, we have to make this a computed property to tell Ember to update our frost-select-dropdown.
+   * Any properties that might be updated after the initial render should be observed by this property.
+   * (@theotherdude 12/20/2017)
+   * @returns {string} 'frost-select-dropdown'
+   */
+  frostSelectDropdownName () {
```
gkchestertron, line 269: OK, I get it... but does it have to be this property? Couldn't we make new computed props that pass along items and selectedItems? I think that would be more sensible than returning a hard-coded string from an otherwise unrelated computed prop just to get them observed.

gkchestertron (approved):

addon/components/frost-select.js
TheOtherDude: John and I talked about this in person. It doesn't seem to matter how we pass the named arguments to the component helper, we only trigger a fresh render when we update the positional argument. 

The tests I added fail without this change.

TheOtherDude:
I discussed alternatives with @notmessenger and @juwara0 . One theory was that using a computed property that returned copies of these named properties might work. I tried it, but unfortunately the tests still failed.
```
  @readOnly
  @computed('items')
  computedItems (items) {
    return items.slice()
  },
```
```

  {{to-elsewhere
    named=renderTarget
    send=(component 'frost-select-dropdown'
      $element=$element
      filter=filter
      hook=hook
      items=computedItems
      multiselect=multiselect
      onClose=(action 'closeDropDown')
      onFilterInput=(action 'filterInput')
      onSelect=(action 'selectItem')
      selectedItems=selectedItems
      wrapLabels=wrapLabels
    )
  }}
```
Doesn't work ^

notmessenger:
@TheOtherDude It's because https://github.com/ciena-frost/ember-frost-core/blob/master/addon/components/frost-select-dropdown.js#L184 is not watching `items.[]`.  This might actually be the entire problem at play here.

TheOtherDude:
I've confirmed that watching `items.[]` makes no difference.

Logging does not even trigger in the computed property when `items` is updated. If I add `{{log computedItems}}` to the template, then I can see it recomputing, but the drop down still does not re-render.

notmessenger:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/530/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

notmessenger (approved):

PR #531, opened 2017-12-21
Add object helper

### LINKED ISSUE #14738: Cannot convert object to primitive value
This is similar to https://github.com/emberjs/ember.js/issues/14187, but I have a reproduction in Ember 2.10 with no addons:

https://ember-twiddle.com/4df2d62bf5f1c05bd74086dc79326020?openFiles=components.my-component.js%2C

The critical ingredient seems to be that I am taking an object created by the `hash` helper, putting it inside an `Array`, and passing the array into another component. When such an array hits `toString` it causes an exception.

(It's hitting `toString` in the first place to support an assertion.)

EWhite613:

# CHANGELOG

* Add object helper to fix https://github.com/emberjs/ember.js/issues/14738

### DIFF
--- a/addon/helpers/object.js
+++ b/addon/helpers/object.js
@@ -1,0 +1,14 @@
+/*
+* ewhite 21/12/2017
+* Fixes https://github.com/emberjs/ember.js/issues/14738
+* Example: https://ember-twiddle.com/2260b80be13a1d6ca5e59c99c40ec51a?openFiles=templates.application.hbs%2C
+* Cause: https://github.com/emberjs/ember.js/blob/v2.12.2/packages/ember-metal/lib/property_set.js#L45
+*/
+import Ember from 'ember'
+const {Helper, merge} = Ember
+
+export function object (_, obj) {
+  return merge({}, obj)
+}
+
+export default Helper.helper(object)
--- a/app/helpers/object.js
+++ b/app/helpers/object.js
@@ -1,0 +1,4 @@
+/**
+ * Re-export the object helper into the app namespace
+ */
+export {default, object} from 'ember-frost-core/helpers/object'
--- a/tests/integration/helpers/object-test.js
+++ b/tests/integration/helpers/object-test.js
@@ -1,0 +1,181 @@
+/**
+ * Integration test for the object-helper component
+ */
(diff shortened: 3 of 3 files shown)


dafortin (approved):

notmessenger (dismissed):
* The dummy app has not been updated to use this new helper
* Rather than creating a new helper that must then be removed once we upgrade to Ember >= 2.13.x I'd prefer to see us modify the `hash` helper instead.

EWhite613:
@notmessenger there's no case for this helper in `ember-frost-core`'s dummy app

As for overriding the hash helper I don't think we can do that as it's not a regiestered Ember helper. Even if you create a hash helper it's still going to use the glimmer hash helper.

~Plus are we going to `2.13.x`?  I expect our next jump is `2.16/17.x` which looks like it still has the issue https://github.com/emberjs/ember.js/blob/v2.16.2/packages/ember-glimmer/lib/helpers/hash.js#L30-L34 (https://github.com/emberjs/ember.js/blob/v2.16.2/packages/ember-metal/lib/property_set.js#L46)~

Edit: nvm looks like they just fixed the Array's to String instead ( https://github.com/emberjs/ember.js/blob/master/packages/ember-utils/tests/to-string-test.js#L24)

I'll see if I can add that fix in (fixing `ember-util` `toString`). Don't know if I'll have access to that tho

notmessenger:
@EWhite613 correct, I was going to point out that the fix was https://github.com/emberjs/ember.js/pull/15117 and was merged into both `2.13.0` and `#lts-2-12` (incorrectly) and that the fix remains all the way to the current versions.

notmessenger:
Even if we can't extend the Ember-provided `hash` helper are we not able to register/provide our own of the same name that gets consumed instead?  I _think_ I did something like that back in the day but I may be misremembering as well.

EDIT: may have to do via an initializer

notmessenger:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/531/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

PR #535, opened 2018-01-15
Fix clear enabled not removing in TextField and TextArea

anhuynh:

# CHANGELOG

* Fix `is-clear-enabled` not being set to false after clicking clear


anhuynh:
## Before

clearenabledtrue

The `_clear` task causes a focus and a trigger input event which calls the `_showClear` task twice. Since `_showClear` was a restartable task, the second call to the task cancels in the middle of the first call and `isClearEnabled` would not get set to false.

## After

clearenabledfalse

Using an enqueue task allows the first `_showClear` call to complete and `isClearEnabled` gets set to false.

package.json
```diff
@@ -89,7 +89,7 @@
     "child-process-promise": "^2.2.1",
     "ember-cli-babel": "^5.1.7",
     "ember-cli-htmlbars": "^1.1.1",
-    "ember-cli-sass": "7.1.1",
+    "ember-cli-sass": "^7.1.1",
```
EWhite613: Want to keep this fixed to `7.1.1`.  Basically, we float stuff we own for the most part, but others we keep fixed

cstolli:
👍 

[Approved with PullApprove](https://pullapprove.com/ciena-frost/ember-frost-core/pull-request/535/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ciena-frost/ember-frost-core)

package.json
notmessenger: @anhuynh @EWhite613  Actually the reason we keep this pinned to `7.1.1` is because version `7.1.1.` includes `ember-cli-babel@^5.1.7` (https://github.com/aexmachina/ember-cli-sass/blob/497ef86a8010d37bdc4ec11342c86bea275fa2cf/package.json#L28) while version `7.1.2` includes `ember-cli-babel@^6.6.0` (https://github.com/aexmachina/ember-cli-sass/blob/b293ec8dea605f81f0cfaae580421cf5a791a1ec/package.json#L28) and this is problematic because

> Ember CLI 2.13.0 introduced a change from using Babel 5.x to 6.x which has introduced issues between the Ember CLI addons and Ember applications world.  Applications or addons running a version of Ember CLI >= 2.13.0 almost always encounter issues consuming addons running Ember CLI <= 2.12.3.  Conversely, applications or addons running a version of Ember CLI <= 2.12.3 almost always encounter issues when consuming addons running Ember CLI >= 2.13.0

tests/integration/components/frost-text-test.js
```diff
@@ -277,6 +277,20 @@ describe(test.label, function () {
       this.$('input').val(),
       'input value cleared'
     ).to.eql('')
+
+    run.later(() => {
+      expect(
+        this.$('.frost-text').hasClass('is-clear-visible'),
```
notmessenger: Should use `hook` (from `ember-hook` instead of `this.$('.frost-text')`

tests/integration/components/frost-text-test.js
```diff
@@ -277,6 +277,20 @@ describe(test.label, function () {
...
+        'class "is-clear-visible" is not set'
+      ).to.equal(false)
+
+      expect(
+        this.$('.frost-text').hasClass('is-clear-enabled'),
```
notmessenger: Should use `hook` (from `ember-hook` instead of `this.$('.frost-text')`

tests/integration/components/frost-text-test.js
```diff
@@ -277,6 +277,20 @@ describe(test.label, function () {
       this.$('input').val(),
       'input value cleared'
     ).to.eql('')
+
+    run.later(() => {
```
notmessenger, line 281: There actually does not appear to be any other test cases in this test suite that verify that the the `is-clear-visible` and `is-clear-enabled` classes are every applied correctly in the state in which clearing the input value is possible.  Therefore these tests would still pass even if the mechanism to apply the classes being tested for was broken.  I'm not saying that this test case is where this should be asserted but please add an additional test case to this test suite asserting the opposite positive assertions of the ones you have added here.

tests/integration/components/frost-textarea-test.js
```diff
@@ -293,6 +293,20 @@ describe(test.label, function () {
       this.$('textarea').val(),
       'input value cleared'
     ).to.eql('')
+
+    run.later(() => {
```
notmessenger, line 297: There actually does not appear to be any other test cases in this test suite that verify that the the `is-clear-visible` and `is-clear-enabled` classes are every applied correctly in the state in which clearing the input value is possible.  Therefore these tests would still pass even if the mechanism to apply the classes being tested for was broken.  I'm not saying that this test case is where this should be asserted but please add an additional test case to this test suite asserting the opposite positive assertions of the ones you have added here.

tests/integration/components/frost-textarea-test.js
```diff
@@ -293,6 +293,20 @@ describe(test.label, function () {
...
+      expect(
+        this.$('.frost-text').hasClass('is-clear-visible'),
```
notmessenger: Should use `hook` (from `ember-hook` instead of `this.$('.frost-text')`

tests/integration/components/frost-textarea-test.js
```diff
@@ -293,6 +293,20 @@ describe(test.label, function () {
...
+        'class "is-clear-visible" is not set'
+      ).to.equal(false)
+
+      expect(
+        this.$('.frost-text').hasClass('is-clear-enabled'),
```
notmessenger: Should use `hook` (from `ember-hook` instead of `this.$('.frost-text')`

addon/components/frost-text.js
```diff
@@ -101,7 +101,7 @@ export default Component.extend(FrostEventsProxyMixin, {
       yield timeout(200) // Duration of the visibility animation
     }
     this.set('isClearEnabled', showClear)
-  }).restartable(),
+  }).enqueue(),
```
notmessenger, line 104: Your PR description says

> The _clear task causes a focus and a trigger input event which calls the _showClear task twice. 

It would be best to correct the `_clear` task so that it does not cause the `_showClear` task to be initiated twice (or maybe it's the `_clear` task being initiated twice).  This PR is fixing the symptom, not the root cause.

addon/components/frost-textarea.js
```diff
@@ -96,7 +96,7 @@ export default Component.extend(FrostEventsProxyMixin, {
       yield timeout(200) // Duration of the visibility animation
     }
     this.set('isClearEnabled', showClear)
-  }).restartable(),
+  }).enqueue(),
```
notmessenger, line 99: Your PR description says

> The _clear task causes a focus and a trigger input event which calls the _showClear task twice. 

It would be best to correct the `_clear` task so that it does not cause the `_showClear` task to be initiated twice (or maybe it's the `_clear` task being initiated twice).  This PR is fixing the symptom, not the root cause.

notmessenger (changes requested):

tests/integration/components/frost-text-test.js
EWhite613: Those tests already exist https://github.com/ciena-frost/ember-frost-core/blob/master/tests/integration/components/frost-text-test.js#L253-L263

notmessenger: Completely missed those.  Thanks!

tests/integration/components/frost-textarea-test.js
notmessenger: I completely missed that such tests already exist at https://github.com/ciena-frost/ember-frost-core/blob/master/tests/integration/components/frost-text-test.js#L253-L263

addon/components/frost-text.js
anhuynh: I just investigated a little bit further and found that it's not just the `_clear` task that causes `_showClear` twice. When the clear button is clicked, a `focusOut` event occurs before the click event. This is the first time that `_showClear` will be called. The click event then happens which calls `_clear`. `_clear` will cause `_showClear` to be triggered twice. We can make it trigger once by removing the [`.focus()` call here](https://github.com/ciena-frost/ember-frost-core/blob/master/addon/components/frost-text.js#L80) since it's not actually needed to clear the field (**edit:** it's needed to keep focus on the field after a clear). However, this still means that `_showClear` will be called for a total of 2 times from the initial `focusOut` event and from `_clear`. I'm not sure if there is a solution to prevent it from being called twice.

notmessenger:
@anhuynh What is the state of this PR?  Is it going to be completed or can it be closed?

anhuynh:
@notmessenger I replied to your initial review comment. As per the comment, I'm not sure there is a solution to prevent `_showClear` from being called twice since it gets called from the `focusOut` event when the user clicks the clear button and then called again during the `_clear` event to actually clear the input field. If you are not happy with my solution then you may close the pr and I will open an issue.

PR #553, opened 2018-03-27
added autocomplete

### LINKED ISSUE #552: Make an autocomplete (similar to frost-select)
# FEATURE REQUEST / CODE CHANGE
## Summary
This feature will provide an autocomplete component which will function similarly to frost-select with a few behaviour and style changes.

## Motivation
* Describe the big picture of your feature to communicate why it should be added.
  * Why are we doing this? This is being pushed down by UX
  * What use cases does it support? Different behaviour and feel for customers
  * What is the expected outcome?

## Detailed design
This component will extend frost-select and will have the following changes:
* It will have a text area to be filled in (frost-select has a dropdown box with a filter/text area)
* it will only open on character input
* it matches based on the start of labels (frost-select match is anywhere)
* Has a new look and feel

## How we teach this

This is a typical autocomplete

## Alternatives

With this change, frost-select will also be able to do match by start of labels.
frost-select and frost-autocomplete will be somewhat interchangeable.

helrac:
# Overview

## Summary
This provides an autocomplete component (much like frost-select, but with different styling and behaviour)

## Issue Number(s)
* Closes #552 

## Screenshots or recordings

screen shot 2018-03-27 at 3 12 12 pm

screen shot 2018-03-27 at 2 59 14 pm

screen shot 2018-03-27 at 3 14 41 pm

# Semver

- [ ] #none#
- [ ] #patch#
- [x] #minor#
- [ ] #major#

Examples:
* **PATCH**
  * changes to demo app (_/tests/dummy/_)
  * added some options for frost-select
    * added filterType
    * added toggle for disabling space bar to open/close menu
  * added frost-autocomplete

# CHANGELOG
* **Added** frost-autocomplete


srowhani (approved):

notmessenger:
This should be a `minor` because it adds new functionality (I have made this change).

@helrac `master` needs to be merged into your branch.

notmessenger (changes requested):
screen shot 2018-03-28 at 7 45 23 am

^ There is extra space around the element that appears when the browser's focus kicks in.

img_0487

^ This is also evident by the fact that this extra space is clickable

screen shot 2018-03-28 at 7 49 01 am

^ It can also be seen when the element is disabled - there is extra gray color outside of the element border.

screen shot 2018-03-28 at 7 45 46 am

^ Is the spacing above the "No items match your filter." text correct?  Seems there should be some more.

screen shot 2018-03-28 at 7 48 07 am

^ Should the drop down box also have a red border like the element when in "error" mode?

img_0488

^ The red border when in "error" mode disappears when the cursor is placed within the field.  Is this expected?

* onChange - local filtering example
** nothing is occuring in regards to the on change handler
* wrapLabels
** indicate which letter should begin typing to experiene this behavior (it is "m")
* what is the intended usage/behaviour of `isLoading`?
* a document needs to be added to https://github.com/ciena-frost/ember-frost-core/tree/master/docs

addon-test-support/frost-autocomplete.js
```diff
@@ -0,0 +1,96 @@
...
+ * @property {String} [focusedItem] - label of focused item
+ * @property {String} [items] - list of item labels present in dropdown
+ * @property {Boolean} [opened=false] - whether or not select is opened
+ * @property {Number} [tabIndex=0] - tab index of root element
+ * @property {String} [text=''] - text in select for describing what is selected
+ */
+
+import {expect} from 'chai'
+import Ember from 'ember'
+const {$, RSVP, run, typeOf} = Ember // eslint-disable-line
```
notmessenger: What error is occurring requiring the need of `// eslint-disable-line`?  It wasn't needed in the _addon/components/frost-autocomplete-dropdown.js_ file

addon-test-support/frost-autocomplete.js
```diff
@@ -0,0 +1,96 @@
...
+import {$hook} from 'ember-hook'
+import wait from 'ember-test-helpers/wait'
+
+const assign = Object.assign || Ember.assign || Ember.merge // eslint-disable-line
```
notmessenger: What error is occurring requiring the need of `// eslint-disable-line`?

addon-test-support/frost-autocomplete.js
```diff
@@ -0,0 +1,96 @@
...
+
+/* eslint-disable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon-test-support/frost-autocomplete.js
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Verify select exists with expected state
+ * @param {jQuery|String} select - name of Ember hook or jQuery instance
+ * @param {FrostSelectState} state - expected select state
+ */
+export function expectWithState (select, state) {
```
notmessenger: Have you tested that this helper works in tests?  There are a lot of references to the `frost-select` class in this code, which is unexpected.

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    const style = [
+      `bottom:${bottom}`,
+      `left:${left}px`,
+      `max-height:${maxHeight}px`,
+      `top:${top}`,
+      `width:${width}px`
+    ]
+      .join(';')
```
notmessenger: Attach this to the `]` on the previous line

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+
+    return Ember.String.htmlSafe(style)
+  },
+
+  @readOnly
+  @computed('focusedIndex', 'items', 'selectedItems')
+  /**
+   * Get render items
+   * @param {Number} focusedIndex - index of focused item
+   * @param {Array<Object>} items - items to render in select dropdown
```
notmessenger: Correct type syntax is `{Object[]}`.  See http://usejsdoc.org/tags-param.html

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+   * @param {Array<Object>} selectedItems - items that are currently selected
```
notmessenger: Correct type syntax is `{Object[]}`.  See http://usejsdoc.org/tags-param.html

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+   * @returns {Array<Object>} render items
```
notmessenger: Correct type syntax is `{Object[]}`.  See http://usejsdoc.org/tags-param.html

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+   */
+  renderItems (focusedIndex, items, selectedItems) {
+    if (!items) {
+      return []
```
notmessenger: The `@returns` says that this method returns an array of objects but this is an empty array.  Either update the code or the return value.

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+        value: get(item, 'value')
+      }
+    })
+  },
+
+  @readOnly
+  @computed('items')
+  /**
+   * Whether or not to show message for when no items are present
+   * @param {Array<Object>} items - items
```
notmessenger: Correct type syntax is `{Object[]}`.  See http://usejsdoc.org/tags-param.html

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+const FPS = 1000 / 60 // Update at 60 frames per second
+const WINDOW_SPACE = 20
+
+export default Component.extend({
+  // == Dependencies ==========================================================
+
+  // == Keyword Properties ====================================================
+
+  layout,
+  tagName: '',
```
notmessenger: This is not future proof (very near future)

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+  @computed('wrapLabels')
+  /**
+   * The class names for the frost-autocomplete drop down
+   * @param {Boolean} wrapLabels - whether or not autocomplete option text should wrap
+   * @returns {string} the class names for the frost-autocomplete drop down
+   */
+  dropdownClassNames (wrapLabels) {
+    const classNames = ['frost-autocomplete-dropdown']
+    if (wrapLabels) {
+      classNames.push('frost-autocomplete-dropdown-wrap-labels')
```
notmessenger: This is not future proof (very near future) nor the way this should be done.  See https://guides.emberjs.com/v3.0.0/components/customizing-a-components-element/#toc_customizing-the-element-s-class

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+    }
+    return classNames.join(' ')
```
notmessenger: This is not future proof (very near future) nor the way this should be done.  See https://guides.emberjs.com/v3.0.0/components/customizing-a-components-element/#toc_customizing-the-element-s-class

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+      return []
+    }
+
+    return items.map((item, index) => {
+      const classNames = ['frost-select-list-item']
+      const value = get(item, 'value')
+      const isSelected = selectedItems.find((item) => item.value === value) !== undefined
+
+      if (isSelected) {
+        classNames.push('frost-select-list-item-selected')
```
notmessenger: This is not future proof (very near future) nor the way this should be done.  See https://guides.emberjs.com/v3.0.0/components/customizing-a-components-element/#toc_customizing-the-element-s-class

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+      }
+
+      if (index === focusedIndex) {
+        classNames.push('frost-select-list-item-focused')
```
notmessenger: This is not future proof (very near future) nor the way this should be done.  See https://guides.emberjs.com/v3.0.0/components/customizing-a-components-element/#toc_customizing-the-element-s-class

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+      }
+
+      return {
+        className: classNames.join(' '),
```
notmessenger: This is not future proof (very near future) nor the way this should be done.  See https://guides.emberjs.com/v3.0.0/components/customizing-a-components-element/#toc_customizing-the-element-s-class

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+        x: height / 2 + top
+      },
+      height,
+      left: offset.left,
+      top,
+      width: $element.width()
+    }
+  },
+
+  /* eslint-disable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+  _handleArrowKey (upArrow) {
```
notmessenger: Where is this being called/used?

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+    const listItemElements = dropdownListElement.querySelectorAll('li')
+
+    Array.from(listItemElements).forEach((li, index) => {
+      $(li)
+        .off('mousedown')
+        .off('mouseenter')
+    })
+  },
+
+  /* eslint-disable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+          const updatedData = trimLongDataInElement(clonedTextElement)
+
+          if (updatedData) {
+            textElement.textContent = updatedData.text
+            textElement.setAttribute('title', updatedData.tooltip)
+          }
+        }
+
+        if (filter) {
+          const pattern = new RegExp(filter.replace(/[.*+?^${}()|[\]\\]/, '\\$&'), 'i')
```
notmessenger: Add a comment indicating the pattern that will satisfy this RegExp

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+          const textWithMatch = textElement.textContent.replace(pattern,
+            "<span class='frost-select-list-item-text-highlight'>$&</span>")
```
notmessenger: Is hard-coded HTML and `innerHTML` the best approach here?  We can't add a new node to an element and apply/remove classes?  This approach has a "smell" to it.

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    this._isUpdating = false
+  }),
+
+  // == DOM Events ============================================================
+
+  // == Lifecycle Hooks =======================================================
+
+  didReceiveAttrs () {
```
notmessenger, line 381: What is all of this code?  Why are we talking about and deprecating properties?

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+    const $element = this.get('$element')
+    let props = {}
+
+    if ($element) {
+      props = merge(props, this._updatePosition($element))
+    }
+
+    const receivedHook = this.get('receivedHook')
+
+    if (receivedHook && receivedHook !== this.get('hook')) {
```
notmessenger: Why?

`receivedHook && `

addon-test-support/frost-autocomplete.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * @typedef {Object} FrostSelectState
```
juwara0: Should this object name not be updated?

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+        if (this.isDestroyed || this.isDestroying) return
+        this._lastInteraction = Date.now()
+
+        if (!this._isUpdating) {
+          this.get('updateTask').perform()
+        }
+      })
+    }
+
+    /* eslint-disable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon-test-support/frost-autocomplete.js
```diff
@@ -0,0 +1,96 @@
...
+import {$hook} from 'ember-hook'
+import wait from 'ember-test-helpers/wait'
+
+const assign = Object.assign || Ember.assign || Ember.merge // eslint-disable-line
+
+/* eslint-disable complexity */
+/**
+ * Verify select exists with expected state
+ * @param {jQuery|String} select - name of Ember hook or jQuery instance
+ * @param {FrostSelectState} state - expected select state
```
juwara0: The name of the param is `FrostSelectState`? This is a test helper for `frost-autocomplete`

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+import layout from '../templates/components/frost-autocomplete'
+import {keyCodes} from '../utils'
+import FrostSelect from './frost-select'
+
+const {$, get, isEmpty, run} = Ember
+const {DOWN_ARROW, UP_ARROW} = keyCodes
+
+export default FrostSelect.extend({
+  layout,
+  classNames: 'frost-select',
```
juwara0: Shouldn't this class name be `frost-autocomplete`?

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,490 @@
...
+          case ENTER:
+            this._handleEnterKey()
+            return
+
+          case ESCAPE:
+            this.get('onClose')()
+            return
+
+          case TAB:
+            this.get('onClose')()
```
notmessenger: Doesn't need a `return` like the others?

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+      _userInput: false,
+      filterType: 'startsWith',
+      isLoading: false,
+      disabled: false,
+      _autofocus: false,
+      disableSpaceToggle: true
+    }
+  },
+
+  /* eslint-disable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+    const newSelectedIndex = (
+      upArrow ? Math.max(0, selectedItemIndex - 1) : Math.min(items.length - 1, selectedItemIndex + 1)
+    )
+
+    if (newSelectedIndex !== selectedItemIndex) {
+      this.set('focusedIndex', newSelectedIndex)
+    }
+  },
+
+  /* eslint-enable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon/components/frost-select-dropdown.js
```diff
@@ -397,6 +399,7 @@ export default Component.extend({
   },
   /* eslint-enable complexity */
 
+  /* eslint-disable complexity */
```
notmessenger: We should not be employing `eslint-disable complexity` in new code.  It should only be used to temporarily address existing code to work around additions or refactors but in new code it should just be written correctly.

addon/components/frost-autocomplete.js
notmessenger: This doesn't come from `FrostSelect` already?  Did you mean for this to be a different value?

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+  init () {
+    this._super(...arguments)
+
+    run.scheduleOnce('afterRender', this, function () {
+      if (this.get('_autofocus') && $(':focus').length === 0) {
+        this.set('_grabfocus', true)
+      }
+    })
+
+    run.scheduleOnce('afterRender', this, function () {
```
notmessenger: Why not combine these two `run.scheduleOnce('afterRender', this, function () {` together?

notmessenger: logic flows should employ only one `return` statement

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+
+  actions: {
+    onKeyDown (event) {
+      if ([DOWN_ARROW, UP_ARROW].indexOf(event.keyCode) !== -1) {
+        this._handleArrowKey(event.keyCode === UP_ARROW)
+      }
+    },
+
+    onKeyPress (event) {
+      this.set('_userInput', true)
```
notmessenger: Combine multiple `.set()` into single `setProperties()`

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+      this.set('opened', true)
+    },
+
+    _selectItem (selectedItem) {
+      this.set('_userInput', false)
```
notmessenger: Combine multiple `.set()` into single `setProperties()`

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,125 @@
...
+      this.set('filter', get(selectedItem, 'label'))
+    },
+
+    onClear () {
+      const props = {
+        opened: false,
+        internalSelectedValue: undefined
+      }
+
+      this.set('_userInput', false)
```
notmessenger: Combine multiple `.set()` into single `setProperties()`

addon/components/frost-select-dropdown.js
```diff
@@ -421,7 +424,12 @@ export default Component.extend({
         }
 
         if (filter) {
-          const pattern = new RegExp(filter.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'gi')
+          let pattern
+          if (this.filterType === 'startsWith') {
+            pattern = new RegExp(filter.replace(/[.*+?^${}()|[\]\\]/, '\\$&'), 'i')
```
notmessenger: Add a comment indicating the pattern that will satisfy this RegExp

addon/components/frost-select-dropdown.js
```diff
@@ -421,7 +424,12 @@ export default Component.extend({
...
+          } else {
+            pattern = new RegExp(filter.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'gi')
```
notmessenger: Add a comment indicating the pattern that will satisfy this RegExp

juwara0 (commented):
I did not see any tests for frost-autocomplete. Nevermind on this comment. I found them.

addon/components/frost-select-dropdown.js
```diff
@@ -421,7 +424,12 @@ export default Component.extend({
         }
 
         if (filter) {
-          const pattern = new RegExp(filter.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'gi')
+          let pattern
+          if (this.filterType === 'startsWith') {
```
notmessenger: Why is this component being modified?  If it is necessary then this change requires the semver of this PR to be a MAJOR, right? since we're changing the behavior of this component?  Or has the default behavior remained in place and only new been added, in which case it can be a MINOR?

addon/components/frost-select.js
```diff
@@ -84,6 +84,8 @@ export default Component.extend({
     onInput: PropTypes.func,
     renderTarget: PropTypes.string,
     role: PropTypes.string,
+    disableSpaceToggle: PropTypes.bool,
```
notmessenger: Why is this component being modified?  If it is necessary then this change requires the semver of this PR to be a MAJOR, right? since we're changing the behavior of this component?  Or has the default behavior remained in place and only new been added, in which case it can be a MINOR?

addon/styles/_frost-autocomplete.scss
```diff
@@ -0,0 +1,95 @@
...
+  height: 80px;
+  margin: auto;
+  margin-top: 25px;
+
+  svg {
+    height: 50px;
+  }
+}
+
+// sass-lint:disable force-element-nesting
```
notmessenger: There's not a way to construct this without needing to use `sass-lint:disable force-element-nesting`?  This is new code - we should not be employing overrides.

addon/templates/components/frost-autocomplete-dropdown.hbs
```diff
@@ -0,0 +1,47 @@
+{{! Template for the frost-autocomplete-dropdown component }}
```
notmessenger: Remove

addon/templates/components/frost-autocomplete-dropdown.hbs
```diff
@@ -0,0 +1,47 @@
...
+<div class='frost-select-overlay' onclick={{onClose}} />
```
notmessenger: This class should be applied via the `className` property on the component.

addon/templates/components/frost-autocomplete.hbs
```diff
@@ -0,0 +1,38 @@
+{{! Template for the frost-select component }}
```
notmessenger: Remove

addon/templates/components/frost-autocomplete.hbs
```diff
@@ -0,0 +1,38 @@
...
+  placeholder=placeholder
+}}
+
+{{#if _opened}}
+  {{to-elsewhere
+    named=renderTarget
+    send=(component 'frost-autocomplete-dropdown'
+      $element=$element
+      filter=filter
+      hook=(concat hook '-dropdown')
```
notmessenger: Change this to `hook=(concat hook 'autocomplete-dropdown')`.  `frost-select` needs to have a similar change made as well so the sub-elements can have unique name scope even though similar elements and naming scheme.

addon/templates/components/frost-autocomplete.hbs
```diff
@@ -0,0 +1,38 @@
+{{! Template for the frost-select component }}
+{{frost-text
+  class='frost-autocomplete-text'
+  ariaActiveDescendant=ariaActiveDescendant
+  ariaAutocomplete='list'
+  ariaExpanded=true
+  ariaMultiselectable=false
+  ariaOwns='frost-select-list'
+  hook=(concat hook '-text')
```
notmessenger: Change this to `hook=(concat hook 'autocomplete-text')`.  `frost-select` needs to have a similar change made as well so the sub-elements can have unique name scope even though similar elements and naming scheme.

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+    afterEach(function () {
+      sandbox.restore()
+    })
+
+    describe('at correct width', function () {
+      const maximumWidth = 330
+      const minimumWidth = 175
+      const specifiedWidth = 500
+
+      it('should if no width is specified, and container is small', function () {
```
notmessenger: It should what?  Modify the grammar of this statement.

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+        this.$().css('width', '100px')
+        const actual = Math.floor(this.$('.frost-select')[0].getBoundingClientRect().width)
+        expect(actual, 'it has the minimum width').to.equal(minimumWidth)
+      })
+
+      it('should if no width is specified, and container is large', function () {
```
notmessenger: It should what?  Modify the grammar of this statement.

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+                keyCode: SPACE
+              })
+            )
+          return wait()
+        })
+
+        it('should render as expected', function () {
+          expect($hook('autocomplete-text-input').text()).to.equal('')
+          expect($hook('autocomplete-dropdown').length).to.equal(0)
+
```
notmessenger: Remove line break

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+          return wait().then(function () {
+            $hook('autocomplete-text-input').focus()
+            return wait()
+          })
+        })
+
+        it('should render as expected', function () {
+          expect($hook('autocomplete-text-input').is(':focus')).to.equal(true)
+          expect($hook('autocomplete-dropdown').length).to.equal(0)
+
```
notmessenger: Remove line break

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+          value: 'Al Simmons'
+        }
+      ]
+      beforeEach(function () {
+        this.set('data', data)
+      })
+
+      it('should render as expected', function () {
+        expect($hook('autocomplete-dropdown').length).to.equal(0)
+
```
notmessenger: Remove line

tests/dummy/app/pods/autocomplete/template.hbs
```diff
@@ -0,0 +1,260 @@
...
+
+  <div class="section-title">Content</div>
+  <hr>
+  <div class="section">
+    <div class="example">
+      <div class="title">data</div>
+      <div class="demo">
+        {{frost-autocomplete
+          data=data
+          hook='mySelect'
```
juwara0: The hooks should be updated in this file since this is autoSelect.

tests/dummy/app/pods/select/template.hbs
```diff
@@ -648,5 +663,45 @@
...
+          filterType='startsWith'
+        }}
+        {{! END-SNIPPET }}
+      </div>
+      <div class='snippet'>
+        {{code-snippet name='filter-type.hbs'}}
+      </div>
+    </div>
+  </div>
+  <div class="section-header">frost-autocomplete</div>
```
juwara0: This should be in the autocomplete dummy example not in select.

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+          return wait().then(function () {
+            $hook('autocomplete-text-input').focus()
+            return wait()
+          })
+        })
+
+        it('should render as expected', function () {
+          expect($hook('autocomplete-text-input').is(':focus')).to.equal(true)
+          expect($hook('autocomplete-dropdown').length).to.equal(0)
+
```
notmessenger: Remove line

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+                keyCode: SPACE
+              })
+            )
+          return wait()
+        })
+
+        it('should render as expected', function () {
+          expect($hook('autocomplete-text-input').text()).to.equal('')
+          expect($hook('autocomplete-dropdown').length).to.equal(0)
+
```
notmessenger: Remove line

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+                $hook('autocomplete-text-input')
+                  .trigger(
+                    $.Event('keydown', {
+                      keyCode: UP_ARROW
+                    })
+                  )
+                return wait()
+              })
+              it('should move up as expected', function () {
+                // debugger
```
notmessenger: Remove

tests/integration/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,538 @@
...
+      })
+
+      describe('click on component', function () {
+        beforeEach(function () {
+          open('autocomplete-text')
+        })
+
+        it('should render as expected', function () {
+          expect($hook('autocomplete-dropdown').length).to.equal(0)
+
```
notmessenger, line 207: Remove line

notmessenger (changes requested):

addon/components/frost-autocomplete.js
helrac: It needs this to inherit a lot of the behaviour from its parent

notmessenger: I believe the question was asked because the expectation was that the `frost-select` component already has a `classNames: 'frost-select'` entry although looking at it it does not.  However, one should be added and the entry in this PR/file either removed or updated to `frost-autocomplete` if need to have specific styling not applicable to the `frost-select` component.

addon/components/frost-autocomplete.js
cstolli: I prefer early returns in a null check case, not super opinionated here.  I find this more readable than using a `let result = whatever` approach

addon/components/frost-select-dropdown.js
helrac: The default behavior is still intact. I called the original behavior 'contains'

addon/components/frost-autocomplete.js
notmessenger: fair enough

addon/components/frost-select-dropdown.js
notmessenger: Was it called `contains` before?  The questions is whether or not a user needs to make any changes to their existing code in order to obtain the current functionality.

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,128 @@
...
+  @computed('opened', '_userInput', 'filter', 'focused')
+  _opened (opened, userInput, filter, focused) {
+    const open = (opened || focused) && userInput && !isEmpty(filter)
+    return open
+  },
+
+  init () {
+    this._super(...arguments)
+
+    run.scheduleOnce('afterRender', this, function () {
```
sophypal: Why is done in `init`? You can't assume that the DOM exist for this component inside an `init`.

sophypal (changes requested):

addon/components/frost-autocomplete.js
notmessenger: It is scheduled in the `afterRender` queue though.

addon/components/frost-select.js
helrac: the default behavior is the same

addon/components/frost-autocomplete.js
sophypal: Doh! You're right.

addon/templates/components/frost-autocomplete.hbs
helrac: I'll go with '-autocompleteText' so it'll like this myAutoComplete-autocompleteText-input for the input

tests/dummy/app/pods/autocomplete/template.hbs
```diff
@@ -0,0 +1,350 @@
...
+
+  <div class="snippet">
+    {{format-markdown "```handlebars
+{{frost-autocomplete-outlet
+  hook='autocompleteOutlet'
+}}
+```"}}
+  </div>
+
+  <p>In the event you want to render different selects in different outlets you may also used named outlets like so:</p>
```
notmessenger: Change "used" to "use"

addon/templates/components/frost-autocomplete.hbs
```diff
@@ -0,0 +1,37 @@
...
+  disabled=disabled
+  autofocus=autofocus
+  placeholder=placeholder
+}}
+
+{{#if _opened}}
+  {{to-elsewhere
+    named=renderTarget
+    send=(component 'frost-autocomplete-dropdown'
+      $element=$element
```
notmessenger: Do not use `$` for property names

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,435 @@
...
+  /**
+   * Properties for this component. Options are expected to be (potentially)
+   * passed in to the component. State properties are *not* expected to be
+   * passed in/overwritten.
+   */
+  propTypes: {
+    // options
+    $element: PropTypes.object.isRequired,
+    filter: PropTypes.string,
+    items: PropTypes.arrayOf(PropTypes.object).isRequired,
```
notmessenger: Do we need to support `Ember.Object` as well?

addon/components/frost-autocomplete-dropdown.js
```diff
@@ -0,0 +1,435 @@
...
+
+      if (index === focusedIndex) {
+        classNames.push('frost-autocomplete-list-item-focused')
+      }
+
+      return {
+        className: classNames.join(' '),
+        label: get(item, 'label'),
+        selected: isSelected,
+        value
```
notmessenger, line 144: Set to `value: get(item, 'value')` and remove line 126 to align with how already assigning `label` a few lines above.

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,316 @@
...
+    }
+    if (this.onClick) {
+      this.onClick(e)
+    }
+  }),
+
+  _onKeyDown: on('keyDown', function (e) {
+    if (
+      [DOWN_ARROW, UP_ARROW].indexOf(e.keyCode) !== -1 &&
+      !this.get('openend')
```
notmessenger: This appears to be spelled incorrectly.   If indeed it is why did no tests catch this?  Write some additional tests to cover this condition as I would not have expected them to pass since referencing an incorrect variable name.

notmessenger (dismissed):
http://localhost:4200/#/autocomplete

* All examples
  * When you press and hold the clear button
  * or when you tab into the field
    * the browser-applied border of the "X" (in latest Chrome) drops below the input element
      * screenshot:
screen shot 2018-04-13 at 3 31 03 pm

* "State" section
  * "error" example
    * red border disappears when the control is given focus
      * is this intentional/desired?
* "Other" section
  * "Spread" example
    * red border is present - assuming an error is occurring
  * "isLoading" example
    * there is no way to experience this behavior because by the time this content is scrolled to the loading has completed.  I'd add a button that you will place into loading state for a second or two.

addon/components/frost-autocomplete.js
helrac: I actually don't need the opened check, the typo actually let it work! good times :D. Thanks

addon/components/frost-autocomplete-dropdown.js
helrac: I'll do the inverse, value is used above

addon/components/frost-autocomplete.js
notmessenger: When you removed the `&& !this.get('openend')` code did you modify a test case as well?  If not then your test coverage is/was still not accurate as either the inclusion or the removal of this code should have caused a test to fail in one of the conditions or the other.

helrac: When i tried it with it spelled correctly, everything worked fine, except the cursor in the input would move left or right depending on the arrow. I could write a test around this to make sure it's not moving the cursor

addon/components/frost-autocomplete-dropdown.js
notmessenger: You can ignore this comment - I did not see `value` being used twice.

addon/components/frost-autocomplete.js
notmessenger: The code was initially spelled incorrectly and things seemed to be working correctly via what I'm guessing was visual inspection only, since no tests were failing (or the tests were also incorrect), correct?

Then you corrected the spelling and things containing to work correctly via what I'm guessing was visual inspection only, since not tests were failing (or the tests were also incorrect), correct?

Since no change in behavior was observed you then removed the `&& !this.get('openend') logic all together then?

When it was spelled incorrectly did you notice this same behavior:

> everything worked fine, except the cursor in the input would move left or right depending on the arrow.

or did this behavior only begin occurring once you spelled it correctly?

And when you say

> except the cursor in the input would move left or right depending on the arrow

are you referring to the behavior of the component when being used or within test cases?


I'd like to take a step back and getting a broader understanding of what is going on.  I would have expected that the mis-spelling to cause a test to fail. Since it did not it either means that there is no test coverage for this condition or that the test is also wrong.  Likewise, when you remove this logic, I would expect tests to fail and since they did not there is either no test coverage for this condition or the test is wrong.  Where I'm really confused though is the introduction of the conversation around the arrow keys and where they fit it in (see my previous question above).  At first glance I would expect the pressing of an arrow key to cause the cursor to move....

tests/dummy/app/pods/autocomplete/template.hbs
```diff
@@ -284,7 +284,7 @@
     </div>
 
     <div class='example'>
-      <div class='title'>Spread</div>
+      <div class='title'>Sprea - (using error state as an example)</div>
```
notmessenger: Change `Sprea` to `Spread`

tests/dummy/app/pods/autocomplete/template.hbs
```diff
@@ -338,9 +338,15 @@
         {{frost-autocomplete
           data=data
           hook='myMultiSelect'
-          isLoading=true
+          isLoading=isLoading
         }}
         {{! END-SNIPPET }}
+        {{frost-button text='Toggle Loading'
```
notmessenger: This button does not toggle the loading state in the demo app

notmessenger:
This is how the error state now displays:

screen shot 2018-04-16 at 1 45 47 pm

It seems we should either make the entire input have a red border or maybe all borders (including the select items) be red?  I would ask UX how they would like to proceed.

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,325 @@
...
+    width: PropTypes.oneOfType([PropTypes.number, PropTypes.null]),
+    renderTarget: PropTypes.string,
+    filter: PropTypes.string,
+    autofocus: PropTypes.bool,
+    debounceInterval: PropTypes.number,
+
+    // state
+    _element: PropTypes.object,
+    focusedIndex: PropTypes.number,
+    _userInput: PropTypes.bool,
```
sophypal: Why have 2 different conventions or private variables?

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,325 @@
...
+  },
+
+  isFilteredItem (value, filterValue) {
+    if (this.filterType === 'startsWith') {
+      return value.startsWith(filterValue)
+    }
+    return value.indexOf(filterValue) !== -1
+  },
+
+  _handleArrowKey (upArrow) {
```
sophypal, line 79: Can you stick with one convention? `_` or without `_`.

sophypal: There's loads of other ones just like this. I'll refrain from adding the same comments to all but assume I mean all of them.

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,325 @@
...
+      this.set('internalSelectedItem', {value: this.selectedValue})
+    }
+  },
+
+  didInsertElement () {
+    this._super(...arguments)
+
+    // We need jQuery instance of components root DOM node to hand off to
+    // dropdown so it can position itself properly relative to the select
+    this.set('_element', this.$())
```
sophypal: So this can be confusing. Every component has `element`. It's not obvious that `_element` is the jQuery instance of it. I would like this to be `$element`.

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,325 @@
...
+    },
+
+    onClear () {
+      this.setProperties({
+        _userInput: false,
+        opened: false,
+        internalSelectedItem: undefined
+      })
+
+      const onClear = this.get('onClear')
```
sophypal: Wait. So you have a function called `onClear` and a property called `onClear`?

addon/components/frost-autocomplete.js
```diff
@@ -0,0 +1,325 @@
...
+
+      if (typeOf(onClear) === 'function') {
+        run.next(() => {
```
sophypal: `run.next` is essentially just a `setTimeout`. You don't gain any protection if the component were destroyed. Either schedule it in a queue or protect it against the component being destroyed.

addon/components/frost-select-dropdown.js
```diff
@@ -286,6 +288,15 @@ export default Component.extend({
     }
   },
 
+  _getRegexPattern (filter) {
+    if (this.filterType === 'startsWith') {
+      // 'b' in bruce banner would only affect the 'b' in bruce
+      return new RegExp('^[ \n\r]*' + filter.replace(/[.*+?^${}()|[\]\\]/, '\\$&'), 'i')
```
sophypal, line 294: Ewww! Why do you need to do it this way?

addon/components/frost-select.js
```diff
@@ -287,6 +289,38 @@ export default Component.extend({
...
+    if (!isMultiselect) {
+      props.filter = ''
+    }
+
+    this.setProperties(props)
+
+    const onChange = this.get('onChange')
+
+    if (typeOf(onChange) === 'function') {
+      run.next(() => {
```
sophypal, line 315: Same comment about `run.next` as above.

package.json
```diff
@@ -100,6 +100,6 @@
     "ember-truth-helpers": "^1.3.0"
   },
   "pr-bumper": {
-    "coverage": 94.22
+    "coverage": 93.17
```
sophypal: I would expect more tests added.

addon/components/frost-autocomplete.js
helrac: the typo'd part always resulted in true. This is why it was working correctly. I just removed the dead logic. It will make sure the dropdown is open using up/down in my case as long as the input has already had actual interaction.

addon/components/frost-autocomplete.js
helrac: @notmessenger asked me to not use $element

notmessenger: Passing `$element` into a component (as a property) is not a pattern we want to encourage, though I understand what @sophypal is referring to.  It seems a more precise name is in order.

package.json
helrac: misses for coverage are around the following type of logic:

if (this.isDestroyed || this.isDestroying) 

Unsure of a proper way to test this.

And also having the dom fire scrolling events while the mouse is over the dropdown.

addon/components/frost-autocomplete.js
helrac: What would be a more appropriate name? jQElement?

tests/unit/components/frost-autocomplete-test.js
```diff
@@ -0,0 +1,48 @@
...
+  describe('runNext', function () {
+    let functionSpy = sinon.spy()
+
+    beforeEach(function () {
+      functionSpy.reset()
+      component._runNext(functionSpy)
+      return wait()
+    })
+
+    it('should be able to runNext', function () {
```
sophypal, line 44: Can you also test that the passed in callback doesn't execute when the component has `isDestroyed` or `isDestroying` set to `true`.

sophypal (approved):

notmessenger (approved):

PR #556, opened 2018-04-12
Add optional isChevronOnlyClickTrigger and labelComponent props to forst-expand

### LINKED ISSUE #555: Add label component prop and click chevron only prop for frost-expand
# FEATURE REQUEST / CODE CHANGE

## Summary
Add two new optional props to the `frost-expand` component: a prop that makes the chevron the only clickable area to expand/collapse and a prop that uses a component in the label instead of just a plain string.

## Motivation
* Describe the big picture of your feature to communicate why it should be added.
  * Why are we doing this?
So that there are additional options to match UX specs when using `frost-expand` to build other components
  * What use cases does it support?
Additional behaviour and aesthetic options for consumers
  * What is the expected outcome?

## Detailed design

The `isChevronOnlyClickTrigger` prop when `true` will make the chevron icon the only clickable area to expand or collapse instead of having the entire label be clickable. The `onClick` action on the label and chevron will be conditional dependent on the `isChevronOnlyClickTrigger` prop.

When the `labelComponent` is passed in with a component, it will be displayed instead of the expand/collapse string labels. An additional `if` statement can be added in the `frost-expand` template to support showing the `labelComponent`.

## How we teach this

* What names and terminology work best for these concepts and why?
`labelComponent` is self explanatory.
`isChevronOnlyClickTrigger` is self explanatory but may be a bit too long.
* How should this feature be introduced and taught to existing users?
It will be added to the dummy app and any documentation for `frost-expand`

## Drawbacks

I don't think there are any since these are optional props

anhuynh:
# Overview

## Summary
Adds two optional props to `frost-expand`:
* `isChevronOnlyClickTrigger` is a flag that allows the consumer to specify that the chevron should be the only clickable area to expand/close.
* `labelComponent` allows the consumer to use a custom component for the label instead of a plain string value.

## Issue Number(s)

* Closes #555 

#### `isChevronOnlyClickTrigger` set to `true`:
chevrononlyclick

#### A custom component passed in as `labelComponent`:
screen shot 2018-04-11 at 8 54 17 pm

## Checklist
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] I have evaluated if the _README.md_ documentation needs to be updated
* [x] I have evaluated if the _/tests/dummy/_ app needs to be modified
* [ ] I have evaluated if DocBlock headers needed to be added or updated
* [x] I have verified that lint and tests pass locally with my changes
* [ ] If a fork of a dependent package had to be made to address the issue this PR closes:
  * [ ] I noted in the fork's _README.md_ the reason the fork was created
  * [ ] I have opened an upstream issue detailing what was deficient about the dependency
  * [ ] I have opened an upstream PR addressing this deficiency
  * [ ] I have opened an issue in this repository to track this PR and schedule the removal of the usage of the fork

# Semver

- [ ] #none#
- [ ] #patch#
- [x] #minor#
- [ ] #major#

# CHANGELOG

* Added `isChevronOnlyClickTrigger` and `labelComponent` props to `frost-expand`


addon/styles/_frost-expand.scss
```diff
@@ -19,14 +19,22 @@ $frost-color-expand-content-text: $frost-color-grey-3 !global;
     }
   }
 
+  .pointer {
```
notmessenger, line 22: Why did this need to be added?  A pointer cursor appears for everything currently at http://ciena-frost.github.io/ember-frost-core/#/expand Is it because of new DOM?

addon/styles/_frost-expand.scss
```diff
@@ -19,14 +19,22 @@ $frost-color-expand-content-text: $frost-color-grey-3 !global;
...
     flex-direction: row;
     flex-shrink: 0;
     align-items: center;
     justify-content: flex-start;
     margin-bottom: 5px;
-    cursor: pointer;
+
+    .frost-expand-label-chevron {
+      height: 30px;
+      cursor: pointer;
```
notmessenger: Leverage SASS and re-use previous definition rather than redefining here as well

addon/templates/components/frost-expand.hbs
```diff
@@ -1,11 +1,17 @@
...
+    {{frost-icon
+      hook=(concat hookPrefix '-label-icon')
+      pack='frost'
+      icon='chevron'
+    }}
+  </div>
   <div class="{{css}}-label-text" data-test={{hook (concat hookPrefix '-label-text')}}>
-    {{#if expanded}}
+    {{#if labelComponent}}
+      {{component labelComponent}}
```
notmessenger: Should we pass hooks through to assist in testing?

notmessenger (changes requested):
Also add documentation requested in https://github.com/ciena-frost/ember-frost-core/issues/555#issuecomment-381148756

addon/styles/_frost-expand.scss
anhuynh: Yes, it's because of the new DOM. When `isChevronOnlyClickTrigger` is set to `true`, the entire label should not have the pointer cursor so this class is added to the label only when the value is `false`.

notmessenger (approved):

PR #564, opened 2018-04-27
fix autocomplete dropdown not re-rendering with new data

### LINKED ISSUE #563: autocomplete not rending dropdown after data has changed
# BUG REPORT

## Summary
The issue is to-elsewhere component (dropdown) is being rendered before the parent (autocomplete) and when the data being passed into autocomplete is changed, the dropdown isn't re-rendering. The dropdown observer for it's items isn't observing in this case

## Expected Behavior
dropdown should handle changes

## Actual Behavior
dropdown will only render when another re-render is forced on it (such as clicking away, and clicking back in to re-open the dropdown)

## Possible Solution
Remove to-elsewhere fixes it

## Steps to Reproduce
Just hook this into some outlet with data that can change

helrac:
# Overview

## Summary
Removing to-elsewhere from autocomplete so the dropdown doesn't render before autocomplete. And to-elsewhere is just not needed.

* Closes #
closes #563 

# Semver

- [ ] #none#
- [x] #patch#
- [ ] #minor#
- [ ] #major#

# CHANGELOG
* **Removed** to-elsewhere usage in autocomplete


tests/integration/components/frost-autocomplete-test.js
```diff
@@ -551,7 +549,6 @@ describe(test.label, function () {
       })
 
       this.render(hbs`
-        {{frost-autocomplete-outlet hook='myWrapOutlet'}}
```
EWhite613, line 554: Can you mark `frost-autocomplete-outlet` as deprecated

notmessenger:
We are going to allow this PR to be submitted as a PATCH, however know that a change like this should normally be a MAJOR.

Because the _addon/components/frost-autocomplete-outlet.js_ file is part of the public API its removal requires a MAJOR.

However because of the inter-dependencies we have amongst our packages on this one the release of this as a MAJOR will require us to release about a dozen other MAJORS of other packages.

So since this a file that had a very specific use and has only been available for a very short period of time, and because we selfishly prefer not to cascade all the other MAJOR releases we will releases this as a patch.

Our apologies to the developer community at large if our selfishness bites you.

notmessenger:
@helrac Does this same issue exist for `frost-select` and does an issue need to be opened to address it?

notmessenger (approved):

EWhite613:
@notmessenger no `frost-select` is fine. It's because autocomplete was passing filter to the `to-elsewhere` causing update issues, whereas `frost-select` does the filter within it's outlet

PR #567, opened 2018-05-10
Remove deprecated xlink attribute from frost-icon

### LINKED ISSUE #554: Since SVG 2, the xlink:href attribute is deprecated in favor of simply href.
# DEPRECATION REPORT

* [x] I have ensured that the issue isn't already reported
* [x] I have confirmed that the issue is reproducible with the latest released version
* [x] I have deleted the **FEATURE REQUEST / CODE CHANGE** section

## Summary
See https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use for additional information

twjohnston:
# Overview

## Summary
SVG xlink attribute is deprecated and looks like it may be causing issues in Chrome in certain situations.

## Issue Number(s)
* Closes #554 

## Checklist
* [ ] I have added tests that prove my fix is effective or that my feature works
* [x] I have evaluated if the _README.md_ documentation needs to be updated
* [x] I have evaluated if the _/tests/dummy/_ app needs to be modified
* [x] I have evaluated if DocBlock headers needed to be added or updated
* [x] I have verified that lint and tests pass locally with my changes

# Semver

- [ ] #none#
- [x] #patch#
- [ ] #minor#
- [ ] #major#

# CHANGELOG
- Remove deprecated xlink attribute from frost-icon


notmessenger (approved):

notmessenger (changes requested):
* `svg4everybody` at _vendor/svg4everybody.min.js_ needs to be updated to (latest) version that supports this change as well.
   * Need to verify that https://github.com/ciena-frost/ember-frost-core/blob/2a6efdd7ab3faf53ad0da119afc3801fedcc8715/docs/frost-icons.md#inline-svg-rendering still works.
   * Need to verify that icons still display.
* There are 4 failing tests

juwara0:
It looks like there is still a problem with `html` attribute support in Safari 11.   https://github.com/jonathantneal/svg4everybody/issues/128

notmessenger:
@juwara0 Agreed though when running this search it looks as if both attributes are supported - https://github.com/jonathantneal/svg4everybody/search?utf8=✓&q=href&type=

I took this to mean, along with the content of the issue you reference, that both attributes are employed to address the Safari issue but maybe not?

EWhite613 (approved):

